PR #18, opened 2015-11-10
add i18n for listofwikis

Reception123:
Not 100% sure it's done like that..


Reception123:
Special page is not in git repo yet, when it is then this can be merged.

Issue #24, opened 2016-01-19
Make a way so that DB's with hyphens can't be submitted


Reception123:
As A LOT of users are putting hyphens in their DB's we have to make a way to not let them submit their request, or find a way to allow hyphens in Miraheze CNAME's (which I don't think is possible)

Reception123:
#25

JohnFLewis:
Hyphens are perfectly legit on a DNS scale.

Southparkfan:
But we don't allow them in dbnames currently. Not DNS related at all.

JohnFLewis:
The issue however was worded as we need to prevent them be becauses it not possibles with regards to DNS. As such I closed it as the original issue was a non-issue.

Southparkfan:
Migrated to https://phabricator.miraheze.org/T153.

PR #25, opened 2016-01-27
prevent requesting name with hyphens

Reception123:
Please comment how to change to be good to merge, or if it will work. (i saw it in a site)

### DIFF
--- a/SpecialRequestWiki.php
+++ b/SpecialRequestWiki.php
@@ -151,3 +151,11 @@ class SpecialRequestWiki extends SpecialPage {
 		$this->getOutput()->addHTML( '<div class="successbox">' . $this->msg( 'requestwiki-success', $idlink )->plain() . '</div>' );
 	}
 }
+
+function has_hyphens($string)
+{
+    return preg_match('/[\-]/', $string);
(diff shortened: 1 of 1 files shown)


SpecialRequestWiki.php
```diff
@@ -151,3 +151,11 @@ function handleRequestWikiFormInput() {
 		$this->getOutput()->addHTML( '<div class="successbox">' . $this->msg( 'requestwiki-success', $idlink )->plain() . '</div>' );
 	}
 }
+
+function has_hyphens($string)
+{
+    return preg_match('/[\-]/', $string);
+}
+
+if(has_hyphens('-'))
```
NDKilla: This '-' should be replaced by the variable that is used to determine the url/DB name. 

That or I don't know what I'm talking about.

NDKilla: I say this because if i understand that function I believe if(has_hyphens('-')) would always return true and thus always throw new exceptions

Reception123:
@Southparkfan

JohnFLewis:
Not necessary, see #24

PR #29, opened 2016-06-27
Update en.json

penguinstyles:


Southparkfan:
Hi, thanks for the improvements! Can you explain why you replaced 'log in' with 'login'?

penguinstyles:
Hi, that is actually a mistake on my end. I was confused as to the proper usage here, log in is actually more appropriate. I've restored the original message. Thanks for pointing that out and making me double check!

Southparkfan:
Thank you very much!

PR #30, opened 2016-07-22
+ Automatic page refresh when wiki request status is changed

penguinstyles:


penguinstyles:
I installed CreateWiki on my personal development server and this fix was working as intended. I just edited the status of one of my [previous wiki requests](https://meta.miraheze.org/wiki/Special:RequestWikiQueue/1332) and it refreshed as intended after one second.

PR #32, opened 2016-08-04
add Orain es.json

Reception123:
Add file from the Orain CreateWiki (es.json)


Reception123:
phpcodechecker.com gives an error so I will wait for another sysadmin to merge.

Reception123:
I think we should rather wait for a Miraheze-specific one, as request-wiki is more important anyways.

PR #44, opened 2017-01-01
direct Request features link on default mainpage

Reception123:
I think maybe some users are having difficulties finding where they can request features so maybe having it directly linked would be a good idea.


Reception123:
Approved by John :)

PR #45, opened 2017-02-25
escape HTML per T98

Reception123:


Southparkfan:
This breaks the PHP syntax, and is not intended escaping.

PR #47, opened 2017-04-17
add cooldown for wiki requests

Reception123:
To prevent spam and people that request the same wiki twice. Not sure if there's a better way to do this and there might be something defined incorrectly.


Reception123:
@NDKilla @Southparkfan this should be reviewed ASAP to prevent further spam on Meta.

labster:
Instead of using $_SESSION, you might want to consider using the `Mediawiki\Session\Session` [class](https://doc.wikimedia.org/mediawiki-core/master/php/classMediaWiki_1_1Session_1_1Session.html).  I think you can access it by doing `$this->getRequest()->getSession()`.

labster:
LGTM. :+1:
The only change I'd consider making is change `last_submit`'s name to something more CreateWiki related (unless trying to tie it to other actions was the original idea?)

PR #53, opened 2017-06-26
T1764 - Attempt to internationalize the approved, declined and in review messages.

penguinstyles:


Reception123:
Looks good. Hopefully it doesn't break anything.

PR #55, opened 2017-07-14
automatic talkpage notifications

The-Voidwalker:
Also includes a checkbox on the createwiki page for if you so desire to notify the user automatically. A separate checkbox handles whether or not to include information about custom domains.


The-Voidwalker:
Any indication of what went wrong?

Reception123:
@The-Voidwalker All I can say it could be related to https://phabricator.miraheze.org/T1104 (only in this case it was worse, it didn't only not appear on Meta RC, but it also didn't appear on the local wiki RC). For the actual notification, I'm not sure why it didn't work. Hopefully once https://phabricator.miraheze.org/T2005 gets done, testing will be easier.

PR #60, opened 2018-03-07
EDIT_NEW can be determined automatically

JohnFLewis:


Reception123:
Just for the record, fixed https://phabricator.miraheze.org/T1104

PR #61, opened 2018-04-03
Dont require the requestwiki user right to request wikis

MacFan4000:


JohnFLewis:
I have more changes than this to make.

PR #62, opened 2018-04-03
Show category in request detail table

MacFan4000:


JohnFLewis:
$res->cw_category returns a none user friendly name which is why I didn't include it.

PR #65, opened 2018-04-16
Fix private to default to false until user selects private

paladox:


SpecialRequestWiki.php
```diff
@@ -82,10 +82,14 @@ protected function getFormFields() {
 
 	public function onSubmit( array $formData ) {
 		$dbname = $formData['subdomain'] . 'wiki';
-		$private = is_null( $formData['private'] ) ? O : 1;
 		$url = $formData['subdomain'] . ".miraheze.org";
 
 		$request = $this->getRequest();
+		
+		$params['private'] = $request->getVal ('private');
```
JohnFLewis: Why are you rewriting form data?

paladox: We are discussing it on irc, but private seems to be the default when requesting new wiki's according to void.

PR #71, opened 2018-04-23
$comment -> $reason

MacFan4000:
Part of this prevented the reason from being placed in the farmer log as before

### DIFF
--- a/SpecialRequestWiki.php
+++ b/SpecialRequestWiki.php
@@ -147,7 +147,7 @@ class SpecialRequestWiki extends FormSpecialPage {
 		$farmerLogEntry = new ManualLogEntry ( 'farmer', 'requestwiki' );
 		$farmerLogEntry->setPerformer( $this->getUser() );
 		$farmerLogEntry->setTarget( $this->getTitle() );
-		$farmerLogEntry->setComment( $comment );
+		$farmerLogEntry->setComment( $reason );
 		$farmerLogEntry->setParameters(
 			array(
 				'4::sitename' => $formData['sitename'],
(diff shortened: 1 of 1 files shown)


SpecialRequestWiki.php
```diff
@@ -147,7 +147,7 @@ public function onSubmit( array $formData ) {
 		$farmerLogEntry = new ManualLogEntry ( 'farmer', 'requestwiki' );
 		$farmerLogEntry->setPerformer( $this->getUser() );
 		$farmerLogEntry->setTarget( $this->getTitle() );
-		$farmerLogEntry->setComment( $comment );
+		$farmerLogEntry->setComment( $reason );
```
paladox: Should be $formData['reason'],

MacFan4000: @paladox I will mention that $reason is used elsewhere in the file

SpecialRequestWiki.php
```diff
@@ -147,7 +147,7 @@ public function onSubmit( array $formData ) {
 		$farmerLogEntry = new ManualLogEntry ( 'farmer', 'requestwiki' );
 		$farmerLogEntry->setPerformer( $this->getUser() );
 		$farmerLogEntry->setTarget( $this->getTitle() );
-		$farmerLogEntry->setComment( $comment );
+		$farmerLogEntry->setComment( 'reason' );
```
paladox: Should be $formData['reason'],

Reception123:
I am unsure about this, but in any case I think "don't fix what's not broken" applies here

SpecialRequestWiki.php
MacFan4000: @JohnFLewis what should it be?

Reception123:
Even though this was removed, the system seems to be working now, so unless we are sure about the correct way to do this, it shouldn't be merged.

SpecialRequestWiki.php
JohnFLewis: $formData['reason']

JohnFLewis:
@Reception123 this is fixing something that is broken. It's a change in behaviour that's quite big in the logs. Unless this is the now preferred behaviour (inadvertently) then the whole setComment line should be removed.

PR #82, opened 2018-07-06
Use setZoneAccess.php script if MirahezeMagic is installed

paladox:


includes/CreateWiki/SpecialCreateWiki.php
```diff
@@ -157,6 +157,18 @@ public function onSubmit( array $formData ) {
...
 		}
 
+		# creates swift container, if for some reason this dosen't create it, to do it manually do:
+		# . /root/swiftExport && swift post <wiki>-mw && swift post -r ".r:*" <wiki>-mw
+		if ( ExtensionRegistry::getInstance()->isLoaded( 'MirahezeMagic' ) ) {
+			$create_image_container = exec( "/usr/bin/php " .
+				"$IP/extensions/MirahezeMagic/maintenance/setZoneAccess.php --wiki " . wfEscapeShellArg( $DBname ) );
+
+			if ( strpos( $create_image_container, 'Swift container failed to be created for' ) ) {
+				wfDebugLog( 'CreateWiki', 'Fail to create container for wiki ' +  wfEscapeShellArg( $DBname ) );
```
JohnFLewis: a bit of consistency here, why +?

i18n/en.json
```diff
@@ -17,6 +17,7 @@
 	"createwiki-error-notalnum": "The database name may not contain non-alphanumeric characters.",
 	"createwiki-error-notlowercase": "The database name may not contain uppercase letters.",
 	"createwiki-error-notsuffixed": "The database name doesn't end in a valid suffix (e.g. 'wiki').",
+	"createwiki-error-swiftcontainernot": "Failed to create swift container for wiki $1",
```
JohnFLewis: why $1? you don't pass a variable and its fairly obvious which wiki it is to the wiki creator.

JohnFLewis: Now you pass a variable.

PR #88, opened 2018-09-16
Introduced wgCreateWikiGlobalWiki config

paladox:


includes/RemoteWiki.php
```diff
@@ -46,7 +46,8 @@ protected static function newFromConds(
 	}
 
 	private function determineCreationDate() {
-		$res = wfGetDB( DB_MASTER )->selectField(
+		global $wgCreateWikiGlobalWiki;
+		$res = wfGetDB( DB_MASTER, [], $wgCreateWikiGlobalWiki )->selectField(
```
JohnFLewis: minor but this should be DB_REPLICA now anyway :)

paladox: Done.

PR #90, opened 2018-09-29
Grant founder rights to the requested wiki author

paladox:


JohnFLewis:
Uh...

This isn’t how miraheze works. There isn’t a founder.

PR #96, opened 2018-11-08
Add "sitename" param to RequestWikiQueue

paladox:


JohnFLewis:
Sitename shouldn’t be searchable.

paladox:
@JohnFLewis then why did you do https://github.com/miraheze/CreateWiki/blob/master/includes/RequestWiki/RequestWikiQueuePager.php#L90 ? :)

PR #98, opened 2018-11-09
add cwPrivate to creationParams

The-Voidwalker:


paladox:
I think we should expose this to the UI too.

PR #113, opened 2019-05-24
Do not throw exception if wiki exists

paladox:
This fixes an issue which should now display that the wiki exists to wiki creators.


JohnFLewis:
This should exception for most cases. Will resolve a more cleaner way for RequestWiki

PR #122, opened 2019-09-24
see phab T4394

examknow:


examknow:
Travis CI build is just a sonar cloud error. It does not mean anything.

extension.json
```diff
@@ -160,7 +160,10 @@
 			"description": "Boolean. Whether to implement front end logic for marking wikis as private.",
 			"public": true,
 			"value": false
-		}
+		},
```
paladox: Needs to be formatted correctly please.

includes/CreateWiki/SpecialCreateWiki.php
```diff
@@ -5,7 +5,11 @@ public function __construct() {
 	}
 
 	protected function getFormFields() {
-		global $wgCreateWikiUseCategories, $wgCreateWikiCategories, $wgCreateWikiUsePrivateWikis, $wgCreateWikiCDBDirectory;
+		$config = MediaWikiServices::getInstance()->getConfigFactory()->makeConfig( 'createwiki' );
+		$UseCategories = $config->get( 'UseCategories' );
```
paladox: Convention for variable names is lowercase to upper case, so for example $useCategories. (And for the rest)

includes/RequestWiki/SpecialRequestWiki.php
```diff
@@ -144,6 +144,8 @@ public function onSubmit( array $formData ) {
 
 		$request = $this->getRequest();
 
+
```
paladox: Remove double new lines please.

includes/RequestWiki/SpecialRequestWiki.php
```diff
@@ -160,7 +162,15 @@ public function onSubmit( array $formData ) {
...
-
+		
+		$dbr = wfGetDB( DB_REPLICA );
+		$res = $dbr->select( 'cw_requests', 'cw_dbname', __METHOD__ );
+		
+		if ( $dbname == $res ) {
+			$out->addHTML( '<div class="errorbox">' .  $this->msg( 'createwiki-error-wikitaken' )->escaped() . '</div>' );
+			wfDebugLog( 'CreateWiki', 'Wiki name already in use. Requested: ' . $dbname );
+			return false;
+					}
```
paladox: Needs to be formatted correctly. And the rest.

JohnFLewis:
I don't like the use of assigning temporary variables for single usage, it's much cleaner and easier to read using the $config->get call instead of assigning it to a variable.

Plus you're opening up unnecessary DB connections in new changes (which shouldn't be happening really in what should be a noop change).

Then you're comparing strings ($dbame) to database objects ($res) which will always return false.

RhinosF1:
has errors, no one working on

PR #139, opened 2020-04-18
Fix renaming wikis

paladox:


includes/WikiManager.php
```diff
@@ -207,6 +205,8 @@ public function rename( string $new ) {
 				]
 			);
 		}
+		
+		$this->recacheJson( $new );
```
JohnFLewis, line 209: Should just be (). The original place is needed

paladox: Thanks, done.

PR #145, opened 2020-06-04
Localization of summaries when creating default main page

soratako:
I'm not confident, so please check the syntax.


i18n/en.json
```diff
@@ -8,6 +8,7 @@
...
+	"createwiki-defaultmainpage-summary": "Create main page"
```
RhinosF1, line 11: Missing a comma

PR #149, opened 2020-06-12
Throw an error when 'subdomain' is requested

examknow:
This will make CreateWiki throw an error when someone requests a wiki with the name 'subdomain'.


examknow:
`5:00:36 <+JohnLewis> Texas: I get the approach, but subdomain.x may have uses in external farms
15:00:45 <+JohnLewis> Using the blacklist would be more effective here`

Closing PR per @JohnFLewis

PR #150, opened 2020-07-03
Don't need to open wfGetDB twice

paladox:
Reduces duplications.

### DIFF
--- a/includes/WikiManager.php
+++ b/includes/WikiManager.php
@@ -54,7 +54,7 @@ class WikiManager {
 
 		$this->dbname = $dbname;
 		$this->dbw = $newDbw;
-		$this->cwdb = wfGetDB( DB_MASTER, [], $wgCreateWikiDatabase );
+		$this->cwdb = $dbw;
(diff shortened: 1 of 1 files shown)


JohnFLewis:
Setting $this->cwdb at the first call would be more efficient

includes/WikiManager.php
```diff
@@ -5,17 +5,16 @@
 
 class WikiManager {
 	private $dbname = null;
-	private $dbw = null;
```
JohnFLewis, line 8: Why?

paladox: It's not used from what i can tell.

JohnFLewis: Yes it is?

paladox: Where? Searching the file i see no more occurrences of $dbw?

JohnFLewis: Line 55

paladox: Oh damn. Ok putting it back.

PR #154, opened 2020-07-15
Add a ChangeDBCluster script

paladox:
Allows you to update multiple wikis and set their db cluster to a different value.

### DIFF
--- a/maintenance/changeDBCluster.php
+++ b/maintenance/changeDBCluster.php
@@ -6,8 +6,8 @@ if ( $IP === false ) {
 }
 require_once "$IP/maintenance/Maintenance.php";
 
-class ChangeDbCluster extends Maintenance {
-	private $dbObj = null;
+class ChangeDBCluster extends Maintenance {
+	private $dbw = null;
 
 	public function __construct() {
 		parent::__construct();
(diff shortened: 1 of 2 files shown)


maintenance/changeDBCluster.php
```diff
@@ -0,0 +1,62 @@
...
+		parent::__construct();
+		$this->addOption( 'file', 'Path to file where the wikinames are store. Must be one wikidb name per line. (Optional, fallsback to current dbname)', false, true );
+		$this->addOption( 'db-cluster', 'Sets the wikis requested to a different db cluster.', true, true );
+	}
+
+	public function execute() {
+		global $wgCreateWikiDatabase, $wgCreateWikiDatabaseClusters, $wgDBname;
+		
+		if ( !$wgCreateWikiDatabaseClusters ) {
+			$this->fatalError( "You have not set \$wgCreateWikiDatabaseClusters." );
```
JohnFLewis, line 22: Is this strictly a requirement? I don't see it being used elsewhere in the script.

JohnFLewis: Also if you used '' instead of "", you'd not need to escape

paladox: Well, no not strictly required i guess.

I'll remove, thanks!

maintenance/changeDBCluster.php
```diff
@@ -0,0 +1,70 @@
...
+			],
+			__METHOD__
+		);
+	}
+
+	private function recacheJson() {
+		global $wgCreateWikiGlobalWiki;
+
+		$cWJ = new CreateWikiJson( $wgCreateWikiGlobalWiki );
+		$cWJ->resetDatabaseList();
```
JohnFLewis, line 64: What about resetWiki as well?

paladox: Thanks, done.

PR #166, opened 2020-11-16
manageInactiveWikis: Change how we get recent changes

paladox:
No longer output to a seperate script


JohnFLewis:
The maintenance script is the preferred way.

paladox:
> The maintenance script is the preferred way.

Why though? It looks cleaner this way and performant too.

JohnFLewis:
Decentralise the code from a maintenance script point. This will move to RemoteWiki soon anyway

PR #167, opened 2020-11-16
Allow users to specific a db object in the constructor

paladox:
This allows to specify a separate db object.


JohnFLewis:
The object should be initiated by name only, no external dbw reference should be used.

PR #174, opened 2020-11-27
Replace "Reason" with "Description"

R4356th:
"Reason" usually confuses most people new to Miraheze. I think replacing it with "Description" should help reduce the confusion and thus also reduce the number of wiki requests which are declined because of having an unclear description.


R4356th:
Looking at the CreateWiki interface, this will apparently also change the text displayed to Wiki creators when approving or declining requests. I think a new message should be added to eliminate this issue.

R4356th:
This should be ready now.

Universal-Omega (changes requested):
Would also need changed at:

https://github.com/miraheze/CreateWiki/blob/98d4a8673f881643fd336c44256e64d757dfb8ec/includes/CreateWiki/SpecialCreateWiki.php#L69
https://github.com/miraheze/CreateWiki/blob/e47d563eeb979eb43a4002713def4948509a801e/includes/RequestWiki/RequestWikiRequestViewer.php#L212

R4356th:
@Universal-Omega, I have updated SpecialCreateWiki.php. As for RequestWikiRequestViewer.php, the whole point of adding a new string is to ensure that the text displayed to Wiki Creators when approving or declining requests is not changed. Are you sure you want me to change that, too? It does not really make much sense to say "Description" in this context, in my opinion. If you do want to change that then instead of so many changes in PHP files, the en.json and qqq.json could simply be updated.

Universal-Omega:
> @Universal-Omega, I have updated SpecialCreateWiki.php. As for RequestWikiRequestViewer.php, the whole point of adding a new string is to ensure that the text displayed to Wiki Creators when approving or declining requests is not changed. Are you sure you want me to change that, too? It does not really make much sense to say "Description" in this context, in my opinion. If you do want to change that then instead of so many changes in PHP files, the en.json and qqq.json could simply be updated.

Well you changed the i18n message key, as such that message will be invalid, so either create a whole new key, or yes, that will need updated

R4356th:
I reversed my changes updating the previous key. It should be untouched.

Universal-Omega:
> I reversed my changes updating the previous key. It should be untouched.

I'm not sure what you mean, it still looks the same to me.

R4356th:
> > I reversed my changes updating the previous key. It should be untouched.
> 
> I'm not sure what you mean, it still looks the same to me. The best bet would be to maintain the Reason message, and add a new description one so reason is maintained where needed, and description is users on Special:CreateWiki/Special:RequestWiki

That is exactly what I am doing. Reason is still there and there is a new message for Description, which is being used in Special:RequestWiki and Special:CreateWiki. I think you should consider taking a look at my changes again.

Universal-Omega:
> > > I reversed my changes updating the previous key. It should be untouched.
> > 
> > 
> > I'm not sure what you mean, it still looks the same to me. The best bet would be to maintain the Reason message, and add a new description one so reason is maintained where needed, and description is users on Special:CreateWiki/Special:RequestWiki
> 
> That is exactly what I am doing. Reason is still there and there is a new message for Description, which is being used in Special:RequestWiki and Special:CreateWiki. I think you should consider taking a look at my changes again.

Oh, I apologize I missread that. Thank you!

Universal-Omega (approved):

Universal-Omega:
This is especially useful to go with my other PR, #175

Universal-Omega:
While this would've been useful if my PR was merged, since it wasn't, then 1) this will only cause the i18n to be needing updated on all languages by translate wiki again, as well as that Reason is actually more relevant here since you are inputting the reason for requesting the wiki, not necessarily the description of it.

R4356th:
@Universal-Omega, while I agree with you I still think this would had decreased the number of invalid wiki requests.

PR #181, opened 2021-01-07
Rename includes/CreateWikiAliases.php to CreateWiki.i18n.alias.php

Universal-Omega:


JohnFLewis:
Personally, I prefer this, than i18n.alias.php

Universal-Omega:
> Personally, I prefer this, than i18n.alias.php

I understand, I was trying to keep consistency with mediawiki/other extensions, as well as here it is easier to find, and therefore translate, but it's up to you. I don't personally mind either way.

JohnFLewis:
> > Personally, I prefer this, than i18n.alias.php
> 
> I understand, I was trying to keep consistency with mediawiki/other extensions, as well as here it is easier to find, and therefore translate, but it's up to you. I don't personally mind either way.

Moving this file wouldn't keep it consistent with 'as here', because this predates your changes today - arguably your changes today break the consistency.

For me, the MediaWiki standard is poor and is not clean. Keeping the files inside includes and with more meaningful and pretty titles such as CreateWikiAlias.php and stuff like CreateWikiMagic.php is better by design.

Universal-Omega:
> > > Personally, I prefer this, than i18n.alias.php
> > 
> > 
> > I understand, I was trying to keep consistency with mediawiki/other extensions, as well as here it is easier to find, and therefore translate, but it's up to you. I don't personally mind either way.
> 
> Moving this file wouldn't keep it consistent with 'as here', because this predates your changes today - arguably your changes today break the consistency.
> 
> For me, the MediaWiki standard is poor and is not clean. Keeping the files inside includes and with more meaningful and pretty titles such as CreateWikiAlias.php and stuff like CreateWikiMagic.php is better by design.

I can't say I don't agree with you. Would you like me to fix the other extensions then that way?

JohnFLewis:
Please. I'd rather be pretty and good designed than consistent with a feature implemented probably 10 years ago.

Universal-Omega:
> Please. I'd rather be pretty and good designed than consistent with a feature implemented probably 10 years ago.

Understood. Doing now. Thanks!

PR #185, opened 2021-01-08
Add check in generateDatabaseList for checking if $allWikis returns non object

paladox:


JohnFLewis:
This is a problem? Returning a value here could be quite dangerous if it does nothing, with appropriately introducing handling of such an error

PR #189, opened 2021-02-24
Fix grammar

Universal-Omega:


dmehus (approved):
:+1: LGTM. This is an improvement in both grammar and clarity.

PR #194, opened 2021-03-01
Add Bio and Purposes fields to Special:CreateWiki

Universal-Omega:


JohnFLewis:
Why? These only have a purpose in RequestWiki

dmehus:
> Why? These only have a purpose in RequestWiki

Well, @Universal-Omega did ask me for my opinion on whether we should add these fields to CreateWiki. I don't think it's absolutely essential as wiki creators/stewards (possibly system administrators?) are the only groups with the technical capability to access CreateWiki, but I _can_ see a use case for a wiki creator/steward (system administrator?) to want to fill in these fields in Special:CreateWiki.

But as I say, it's not absolutely essential, but would be helpful on rare occasions

JohnFLewis:
But this data exists only in cw_requests, CreateWiki doesn’t interact with that data, nor does it insert data into that table. If you enter the data into these fields, it’s serves no purpose except to make the wiki creator fill out e train fields that are stored no where. Plus 99% of creations don’t even go through this special page

dmehus:
> If you enter the data into these fields, it’s serves no purpose except to make the wiki creator fill out e train fields that are stored no where

@JohnFLewis, oh okay, fair enough. That's a good point then. Yeah, you're right, if we enter this information, it's not stored in RequestWiki, as there's no RequestWiki record. I should've thought of that actually.

PR #195, opened 2021-03-01
Allow $wgCreateWikiBlacklistedSubdomains to be an array

Universal-Omega:
So it's easier to add new regex blacklists, and doesn't have one huge string. Feel free to close this if it isn't a good idea to do though.

### DIFF
--- a/extension.json
+++ b/extension.json
@@ -92,7 +92,7 @@
 	},
 	"config": {
 		"CreateWikiBlacklistedSubdomains": {
-			"description": "String. Regex to match for a blacklisted subdomain.",
+			"description": "String or array. Regex to match for a blacklisted subdomain.",
 			"public": true,
 			"value": "/ /"
 		},
--- a/includes/RequestWiki/SpecialRequestWiki.php
+++ b/includes/RequestWiki/SpecialRequestWiki.php
@@ -105,6 +105,12 @@ class SpecialRequestWiki extends FormSpecialPage {
 	public function onSubmit( array $formData ) {
 		$subdomain = strtolower( $formData['subdomain'] );
 
+		if ( is_array( $this->config->get( 'CreateWikiBlacklistedSubdomains' ) ) ) {
+			$subdomainBlacklist = '/^(' . implode( '|', $this->config->get( 'CreateWikiBlacklistedSubdomains' ) ) . ')+$/';
+		} else {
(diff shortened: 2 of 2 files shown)


dmehus:
@Universal-Omega Was this variable currently a string, and this PR proposes to convert it to an array? If _so_, this seems to make a lot of sense, and simplifies the list of blacklisted CreateWiki subdomains, I think

Universal-Omega:
> @Universal-Omega Was this variable currently a string, and this PR proposes to convert it to an array? If _so_, this seems to make a lot of sense, and simplifies the list of blacklisted CreateWiki subdomains, I think

Yes, it's currently a string, this PR allows it to continue being a string, but also supports it being an array, for simplicity and making it not be just a huge string.

dmehus:
> > @Universal-Omega Was this variable currently a string, and this PR proposes to convert it to an array? If _so_, this seems to make a lot of sense, and simplifies the list of blacklisted CreateWiki subdomains, I think
> 
> Yes, it's currently a string, this PR allows it to continue being a string, but also supports it being an array, for simplicity and making it not be just a huge string.

Oh, gotcha. It simplifies the string then? Well I definitely support moving it to an array.

Universal-Omega:
> > > @Universal-Omega Was this variable currently a string, and this PR proposes to convert it to an array? If _so_, this seems to make a lot of sense, and simplifies the list of blacklisted CreateWiki subdomains, I think
> > 
> > 
> > Yes, it's currently a string, this PR allows it to continue being a string, but also supports it being an array, for simplicity and making it not be just a huge string.
> 
> Oh, gotcha. It simplifies the string then? Well I definitely support moving it to an array.

No it doesn't simplify string, it simplifies config by supporting it as an array. But maintains compatibility with the former string.

dmehus:
> > > > @Universal-Omega Was this variable currently a string, and this PR proposes to convert it to an array? If _so_, this seems to make a lot of sense, and simplifies the list of blacklisted CreateWiki subdomains, I think
> > > 
> > > 
> > > Yes, it's currently a string, this PR allows it to continue being a string, but also supports it being an array, for simplicity and making it not be just a huge string.
> > 
> > 
> > Oh, gotcha. It simplifies the string then? Well I definitely support moving it to an array.
> 
> No it doesn't simplify string, it simplifies config by supporting it as an array. But maintains compatibility with the former string.

Okay, thanks for clarifying.

PR #198, opened 2021-03-08
Add 'name' to bio and purpose fields for consistency

Universal-Omega:


JohnFLewis:
Name probably no longer serves a purpose anymore

Universal-Omega:
> Name probably no longer serves a purpose anymore

Should they be removed from others then?

PR #201, opened 2021-03-10
Re-add echo notifications for declined requests

Universal-Omega:
Changes the behavior from 06a282b, which completely removed the declined notifications type from sending, this in turn, remains to keep it from sending 2 notifications for declined requests while still supporting the `request-declined` type.

Additionally, Currently, it seems like echo does not send *any* email notifications for declined requests or request comments (reliably) - regardless of email preferences set in Special:Preferences, which is something else that needs looked into also. It does still send web notifications though, but even for declined notifications, it sends that it received a comment, not that it was declined, which is what this fixes.


dmehus (commented):
<s>What's this commit doing, @Universal-Omega? Can you please clarify in a sentence or two what it is doing? Thanks.</s>

Answered with @Universal-Omega's edited PR above.

PR #202, opened 2021-03-11
Set setSchemaVars within WikiManager->create

paladox:
This is so wgDBTableOptions gets replaced in the sql files.

### DIFF
--- a/includes/WikiManager.php
+++ b/includes/WikiManager.php
@@ -90,7 +90,7 @@ class WikiManager {
 			return $checkErrors;
 		}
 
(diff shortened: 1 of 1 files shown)


Universal-Omega:
To note I think this was for T4425. I can't remember exactly why this wasn't yet merged though.

Universal-Omega:
Is this still planned to be merged with T4425 being closed?

Universal-Omega:
Closing since this was originally for T4425, and without doing that I'm not sure this is a great idea. Apologies if I'm wrong in closing this.

PR #204, opened 2021-03-14
Fix "Undefined property: RequestWikiAIJob::$description"

paladox:


dmehus (commented):
:+1:

includes/RequestWiki/RequestWikiAIJob.php
```diff
@@ -17,7 +17,7 @@ public function run() {
 		if ( file_exists( $modelFile ) ) {
 			$modelManager = new ModelManager();
 			$pipeline = $modelManager->restoreFromFile( $modelFile );
-			$tokenDescription = (array)strtolower( $this->description );
+			$tokenDescription = (array)strtolower( $this->params['description'] );
```
Universal-Omega, line 20: Why not just `$wr->description`?

PR #215, opened 2021-05-06
Use textarea type for reason and comment fields

Universal-Omega:
This has been mentioned multiple times so doing this. It also adds a help message to the reason field.

Also makes it so you can't post an empty comment or open a wiki request with a completely empty reason (or contains only whitespace or newlines)

### DIFF
--- a/i18n/en.json
+++ b/i18n/en.json
@@ -22,7 +22,7 @@
 	"createwiki-error-notsuffixed": "The database name doesn't end in a valid suffix (e.g. 'wiki').",
 	"createwiki-error-swiftcontainernot": "Failed to create swift container for wiki $1",
 	"createwiki-error-usernotcreated": "The requester user was not created on the wiki.",
-	"createwiki-help-reason": "Enter 2-4 sentences describing the topic, scope, and purpose of the wiki you're requesting.",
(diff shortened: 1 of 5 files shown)


dmehus:
Yeah, I think I've even suggested this in the past year (last fall sometime, as I recall). I'd suggest a 3-4 line text area field. How many lines have you proposed with https://github.com/miraheze/CreateWiki/pull/215/commits/fe11a70d7ccf6f23d331f229d704936dc325f6ac? No concerns otherwise, so 👍

Universal-Omega:
> Yeah, I think I've even suggested this in the past year (last fall sometime, as I recall). I'd suggest a 3-4 line text area field. How many lines have you proposed with [fe11a70](https://github.com/miraheze/CreateWiki/commit/fe11a70d7ccf6f23d331f229d704936dc325f6ac)? No concerns otherwise, so 👍

Default is 25 rows, I just fixed it to be a 6 row field

dmehus:
I think @Universal-Omega changed it to 4 rows per discussion on IRC, so LGTM and 👍 per that. Feel free to do the `i18n` messages; if any changes, I can always modify them after

dmehus (approved):
This seems fine, and you used the correct `you're`, so 👍

PR #242, opened 2021-10-01
Cache extension list instead of loading each page load

Universal-Omega:
If this doesn't work then the entire change in mw-config at least should be reverted.


RhinosF1 (approved):
EMERGENCY

PR #254, opened 2021-10-26
CWJ: add experimental to dblists

RhinosF1:
will allow some manipulation in future if multiversion does come & for running maint scripts only on experimental wikis

### DIFF
--- a/includes/CreateWikiJson.php
+++ b/includes/CreateWikiJson.php
@@ -102,13 +102,13 @@ class CreateWikiJson {
 				}
 			}
 			if ( $wiki->wiki_experimental == 1 ) {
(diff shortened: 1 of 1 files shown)


Universal-Omega (dismissed):

Universal-Omega:
I would like to note that I've just been thinking about this and I don't necessarily see a use-case for it to be added to current dblists. A maintenance script similar to the one that generates database lists for extensions, can be created for this. While doing this is for uses like https://github.com/miraheze/CreateWiki/blob/d7241ec/includes/WikiInitialise.php#L62. Which that doesn't necessarily apply to experimental.

PR #258, opened 2021-11-05
New decline icon

Universal-Omega:


RhinosF1 (commented):
Is there nothing good in OOUI?

Universal-Omega:
> Is there nothing good in OOUI?

As mentioned on IRC that is a good point. Will consider in future, but this should still be done as an intermediate fix, as the current icon is far to small.

PR #266, opened 2021-12-09
Clarify wording on RequestWiki form

AgentIsai:
Clarify wording on Special:RequestWiki to change "Enter a sufficient summary" to "Enter a summary of at least 2-3 sentences" to match the "Needs more details" canned response which makes mention of entering a 2-3 sentence description.

### DIFF
--- a/i18n/en.json
+++ b/i18n/en.json
@@ -19,7 +19,7 @@
 	"createwiki-error-notalnum": "The database name may not contain non-alphanumeric characters.",
 	"createwiki-error-notlowercase": "The database name may not contain uppercase letters.",
 	"createwiki-error-notsuffixed": "The database name doesn't end in a valid suffix (e.g. 'wiki').",
-	"createwiki-help-reason": "Please enter a sufficient summary describing the topic, scope, and purpose of the wiki you're requesting.",
+	"createwiki-help-reason": "Please enter a summary of at least 2-3 sentences describing the topic, scope, and purpose of the wiki you're requesting. If you don't, your request will likely be declined.",
 	"createwiki-label-category": "Category:",
 	"createwiki-label-dbname": "Database name:",
 	"createwiki-label-language": "Language:",


Universal-Omega:
I don't really like this change personally. One sentence is fine in many cases.

Reception123:
I agree, I think 2-3 sentences is better as it provides a more specific approximation for users to know about how much they should write.

Universal-Omega:
As for the canned response being that, typically if a wiki is declined for that the topic requires more detailed information. That isn't required for all topics, if it's clearly going to adhere to policies then just once sentence might be OK.

Universal-Omega:
> I agree with @Universal-Omega, I think 2-3 sentences is better as it provides a more specific approximation for users to know about how much they should write.

I was actually saying the opposite of that. There really isn't a common approximation that covers all topics, 2-3 sentences is for some, but for others just one would be OK.

AgentIsai:
> I don't really like this change personally. One sentence is fine in many cases. When I created this help message I was going to say that but didn't intentional as I don't really like the wording. As for the last part, that justly sounds a bit direct to me, and isn't always true. It's a wiki creator's perogative to decline or approve based on their opinion of the request. I personally think it's fine how it is.

Completely true, one sentence is fine but generally only for private wikis. One sentence (unless it's a run on sentence) is seldomly enough to describe a wiki's scope and to give a wiki creator an good idea of whether the wiki will follow the Content Policy or be problematic. The current system message says "Enter a sufficient summary" which isn't very descriptive. Some requesters may think a simple summary saying "Wiki about X thing." is sufficient when in reality, it's not as we literally have no idea what their wiki will cover, might not know what the "X thing" is, whether they seem to understand our policies, etc. By giving a reference point of "2-3 sentences," requesters know that their request shouldn't be too short and indescriptive like that and requesters might potentially end up saving time and having their requests approved the first time around instead of having to go and edit their request if it gets declined (if they even know how to, that is). Of course, as you said, wiki creators do hold a prerogative and will continue to hold it, this is not a policy change or something set in stone. Wiki creators are free to continue to exercise their discretion and approve wiki requests with less than 2-3 sentences as their description, including private wiki requests, if they believe, in their judgement, that the request seems good.

Universal-Omega:
Closing per our conversation on Discord

PR #268, opened 2021-12-20
Add realm option to setVariables to set SiteConfiguration tags

Universal-Omega:


RhinosF1:
Merging so it gets cut before the weekly PRs

PR #280, opened 2022-01-27
Rename two configuation variables

Universal-Omega:


Universal-Omega:
> This can't be deployed without either causing a security issue or relying on the person deploying to have a fallback set.
> 
> This should set the fallback and accept either for at least 1 commit.

I already have a PR to mw-config to set this new config, in addition to the old one to prevent a lapse of no blacklist at all.

PR #283, opened 2022-01-28
Convert to PSR-4 and use AutoloadNamespaces

lens0021:
Caution:
- (Obsoleted) This changes the names of classes, thus there should be a guarantee that any of them is not referenced by any other extensions. I couldn't find it in my rough search for ManageWiki and WikiDiscover, but I can't be sure.
- There are no unit tests on this extension, so this could break something even if the CI is passed.


Universal-Omega:
This can't be done yet. CreateWiki classes are used in ManageWiki, GlobalNewFiles, MirahezeMagic. maybe others as well.

lens0021:
> This can't be done yet. CreateWiki classes are used in ManageWiki, GlobalNewFiles, MirahezeMagic. maybe others as well.

Then, we can use `class_alias()`. An example code's here: https://gerrit.wikimedia.org/g/mediawiki/core/+/e1e4f7514fef71e747c1f59ccce8badba6a079b3/includes/block/DatabaseBlock.php#1207

Universal-Omega:
> > This can't be done yet. CreateWiki classes are used in ManageWiki, GlobalNewFiles, MirahezeMagic. maybe others as well.
> 
> Then, we can use `class_alias()`. An example code's here: https://gerrit.wikimedia.org/g/mediawiki/core/+/e1e4f7514fef71e747c1f59ccce8badba6a079b3/includes/block/DatabaseBlock.php#1207

yeah that would be best for now.

Universal-Omega:
WikiInitialise shouldn't be moved namespaces. It must remain where it is due to the configuration setup. I don't even think we should alias that one. Additionally it shouldn't be autoloaded so shouldn't be namespaced.

.phan/config.php
```diff
@@ -14,9 +14,11 @@
 	]
 );
 
-$cfg['suppress_issue_types'] = [
-	'PhanTypeArraySuspiciousNullable',
-	'SecurityCheck-LikelyFalsePositive',
-];
+$cfg['suppress_issue_types'] = array_merge(
```
Universal-Omega, line 17: This shouldn't be done (unless theres a reason for this?). It was intentionally explicitly specified so that it picks up on deprecations, etc..

Universal-Omega (changes requested):

.phan/config.php
lens0021: I don't know why, but methods of User are considered as internal before including them.

<details>
<summary>[debug] Restarted process exited 1</summary>

```
includes/CreateWiki/CreateWikiJob.php:24 PhanAccessMethodInternal Cannot access internal method \User::newSystemUser() of namespace \ defined at /var/www/html/w/includes/user/User.php:809 from namespace \MediaWiki\Extension\CreateWiki\CreateWiki
includes/CreateWiki/CreateWikiJob.php:40 PhanAccessMethodInternal Cannot access internal method \User::newSystemUser() of namespace \ defined at /var/www/html/w/includes/user/User.php:809 from namespace \MediaWiki\Extension\CreateWiki\CreateWiki
includes/CreateWiki/CreateWikiJob.php:41 PhanAccessMethodInternal Cannot access internal method \User::newSystemUser() of namespace \ defined at /var/www/html/w/includes/user/User.php:809 from namespace \MediaWiki\Extension\CreateWiki\CreateWiki
includes/CreateWiki/CreateWikiJob.php:46 PhanAccessMethodInternal Cannot access internal method \User::newSystemUser() of namespace \ defined at /var/www/html/w/includes/user/User.php:809 from namespace \MediaWiki\Extension\CreateWiki\CreateWiki
includes/CreateWiki/SpecialCreateWiki.php:97 PhanAccessMethodInternal Cannot access internal method \User::getName() of namespace \ defined at /var/www/html/w/includes/user/User.php:2115 from namespace \MediaWiki\Extension\CreateWiki\CreateWiki
includes/RequestWiki/RequestWikiAIJob.php:33 PhanAccessMethodInternal Cannot access internal method \User::newSystemUser() of namespace \ defined at /var/www/html/w/includes/user/User.php:809 from namespace \MediaWiki\Extension\CreateWiki\RequestWiki
includes/RequestWiki/RequestWikiAIJob.php:36 PhanAccessMethodInternal Cannot access internal method \User::newSystemUser() of namespace \ defined at /var/www/html/w/includes/user/User.php:809 from namespace \MediaWiki\Extension\CreateWiki\RequestWiki
includes/RequestWiki/RequestWikiQueuePager.php:72 PhanAccessMethodInternal Cannot access internal method \User::getName() of namespace \ defined at /var/www/html/w/includes/user/User.php:2115 from namespace \MediaWiki\Extension\CreateWiki\RequestWiki
includes/RequestWiki/RequestWikiQueuePager.php:124 PhanAccessMethodInternal Cannot access internal method \User::getId() of namespace \ defined at /var/www/html/w/includes/user/User.php:2082 from namespace \MediaWiki\Extension\CreateWiki\RequestWiki
includes/RequestWiki/RequestWikiRequestViewer.php:78 PhanAccessMethodInternal Cannot access internal method \User::getId() of namespace \ defined at /var/www/html/w/includes/user/User.php:2082 from namespace \MediaWiki\Extension\CreateWiki\RequestWiki
includes/RequestWiki/RequestWikiRequestViewer.php:78 PhanAccessMethodInternal Cannot access internal method \User::getName() of namespace \ defined at /var/www/html/w/includes/user/User.php:2115 from namespace \MediaWiki\Extension\CreateWiki\RequestWiki
includes/RequestWiki/RequestWikiRequestViewer.php:112 PhanAccessMethodInternal Cannot access internal method \User::getName() of namespace \ defined at /var/www/html/w/includes/user/User.php:2115 from namespace \MediaWiki\Extension\CreateWiki\RequestWiki
includes/RequestWiki/RequestWikiRequestViewer.php:117 PhanAccessMethodInternal Cannot access internal method \User::getId() of namespace \ defined at /var/www/html/w/includes/user/User.php:2082 from namespace \MediaWiki\Extension\CreateWiki\RequestWiki
includes/RequestWiki/SpecialRequestWiki.php:34 PhanAccessMethodInternal Cannot access internal method \User::isRegistered() of namespace \ defined at /var/www/html/w/includes/user/User.php:2965 from namespace \MediaWiki\Extension\CreateWiki\RequestWiki
includes/RequestWiki/WikiRequest.php:115 PhanAccessMethodInternal Cannot access internal method \User::getId() of namespace \ defined at /var/www/html/w/includes/user/User.php:2082 from namespace \MediaWiki\Extension\CreateWiki\RequestWiki
includes/RequestWiki/WikiRequest.php:121 PhanAccessMethodInternal Cannot access internal method \User::getId() of namespace \ defined at /var/www/html/w/includes/user/User.php:2082 from namespace \MediaWiki\Extension\CreateWiki\RequestWiki
includes/RequestWiki/WikiRequest.php:166 PhanAccessMethodInternal Cannot access internal method \User::getName() of namespace \ defined at /var/www/html/w/includes/user/User.php:2115 from namespace \MediaWiki\Extension\CreateWiki\RequestWiki
includes/RequestWiki/WikiRequest.php:167 PhanAccessMethodInternal Cannot access internal method \User::getName() of namespace \ defined at /var/www/html/w/includes/user/User.php:2115 from namespace \MediaWiki\Extension\CreateWiki\RequestWiki
includes/RequestWiki/WikiRequest.php:187 PhanAccessMethodInternal Cannot access internal method \User::getName() of namespace \ defined at /var/www/html/w/includes/user/User.php:2115 from namespace \MediaWiki\Extension\CreateWiki\RequestWiki
includes/RequestWiki/WikiRequest.php:208 PhanAccessMethodInternal Cannot access internal method \User::getId() of namespace \ defined at /var/www/html/w/includes/user/User.php:2082 from namespace \MediaWiki\Extension\CreateWiki\RequestWiki
includes/RequestWiki/WikiRequest.php:209 PhanAccessMethodInternal Cannot access internal method \User::getId() of namespace \ defined at /var/www/html/w/includes/user/User.php:2082 from namespace \MediaWiki\Extension\CreateWiki\RequestWiki
includes/RequestWiki/WikiRequest.php:293 PhanAccessMethodInternal Cannot access internal method \User::getId() of namespace \ defined at /var/www/html/w/includes/user/User.php:2082 from namespace \MediaWiki\Extension\CreateWiki\RequestWiki
```

</details>

Universal-Omega: Because we are namespacing now, it isn't in the main (`\`) namespace so it isn't part of `@internal` anymore. Can you explicitly ignore the internal ones then?

Universal-Omega: Or maybe even phan file-wide suppresions for them? Since its only two files.

includes/Hooks.php
```diff
@@ -163,7 +167,7 @@ public static function onBeforeCreateEchoEvent(
 			'group' => 'positive',
 			'section' => 'alert',
 			'canNotifyAgent' => true,
-			'presentation-model' => EchoCreateWikiPresentationModel::class,
+			'presentation-model' => \MediaWiki\Extension\CreateWiki\Notifications\EchoRequestCommentPresentationModel::class,
```
Universal-Omega, line 170: Can we add `use` statements for this and the rest?

Universal-Omega (changes requested):

.phan/config.php
lens0021: It looks like seven files to me and it's too many. I will add PhanAccessMethodInternal to config.php.

Universal-Omega: Sounds good! Thanks!

includes/Hooks.php
lens0021: Yes, I will. I just didn't because they are used only once each.

Can

https://github.com/lens0021/CreateWiki/blob/d81f4eb914170472ac81a3edb0ea5e57ffe3ceda/includes/CreateWikiOOUIForm.php#L20
```php
new \OOUI\PanelLayout();
```

be same?

Universal-Omega: The OOUI I'm indifferent about, as it wasn't before. It's entirely necessary, but if you want to I guess it's fine.

lens0021: Ok, I'll leave OOUI as they are because I think it's simpler.

Universal-Omega:
I'm just now thinking about this. Very sorry I didn't sooner, but maybe we should namespaces under `Miraheze\CreateWiki`. That's what I used for the RemovePII extension when first developing it because it's shorter, cleaner, and works for Miraheze extensions, but still doesn't make it to Miraheze specific in my mind. If you do want to use what it is now, that'll probably be OK though.

lens0021:
If `Miraheze\<ExtensionName>` will be a standard for extensions by Miraheze, I am fine with it because I think it is more important to have a constant convention than merging my PR.
One thing to note is the Best practices for extensions[1] have a clause for namespaces, though the document is a draft, and the discussion for it has been started recently.[2]

1: https://www.mediawiki.org/wiki/Special:MyLanguage/Best_practices_for_extensions
2: https://lists.wikimedia.org/hyperkitty/list/wikitech-l@lists.wikimedia.org/thread/5EURNXRANAEU3MC4OKSDBXIIBUMKGNNZ/

Universal-Omega:
When I did RemovePII I did intend to make Miraheze\<ExtensionName> the standard for all our extensions and still believe it's preferable. I planned to do the rest of the extensions as well but never got to it. Thank you!

Universal-Omega (approved):

Universal-Omega:
I will test this tomorrow. Otherwise all looks good.

Universal-Omega:
Is it necessary to use AutoloadNamespaces for maintenance here, which loads all maintenance scripts, which is a behavioural change from before which only loaded some?

lens0021:
I just did what I saw in other extensions. For instance, AbuseFilter do that. https://github.com/wikimedia/mediawiki-extensions-AbuseFilter/blob/a55c0e51ed6d714dd922ec21a20f0ad356076d7f/extension.json#L194
Honestly, I don't well understand the pros of PSR-4. (sorry) So if you profer with it, I would revert the maintenance part.

Universal-Omega:
> I just did what I saw in other extensions. For instance, AbuseFilter do that. https://github.com/wikimedia/mediawiki-extensions-AbuseFilter/blob/a55c0e51ed6d714dd922ec21a20f0ad356076d7f/extension.json#L194 Honestly, I don't well understand the pros of PSR-4. (sorry) So if you profer with it, I would revert the maintenance part.

I think the namespacing should be kept, but I don't really think they need autoloader. But I'm not sure, is there a disadvantage of loading many classes (performance I mean), if not then they can be, if there is, then I'd just rather reduce the number of loaded classes is the only reason I brought it up.

Universal-Omega (approved):

RhinosF1:
@lens0021 this has a conflict

RhinosF1:
Can this be merged? What's left here?

RhinosF1 (commented):
Can class_alias only be a soft deprecation or can we emit a warning?

lens0021:
> Can this be merged? What's left here?

Universal-Omega mentioned the test, but I have no idea what is.

> Can class_alias only be a soft deprecation or can we emit a warning?

https://phabricator.wikimedia.org/T293144 says:

> It is not really possible to hard deprecate class aliases, …

So there is no implementation yet?

RhinosF1:
> > Can this be merged? What's left here?
> 
> Universal-Omega mentioned the test, but I have no idea what is.

Assigning to him then
> > Can class_alias only be a soft deprecation or can we emit a warning?
>
> https://phabricator.wikimedia.org/T293144 says:
> 
> > It is not really possible to hard deprecate class aliases, …
> 
> So there is no implementation yet?

Thanks!

Universal-Omega (changes requested):
I unfortunately created a conflict with #301. Mind resolving it? Once done, this can hopefully be merged. WikiManagerTest can also be namespaced, and use TestAutoloadNamespaces. It is possible that additional conflicts could be later created with #297 if that ends up being merged before this. I am a bit hesitant to merge this at the moment due to the size, but probably can be done soon. I might try and just get this done first to avoid further conflicts of it. Even if it postpones others, but it can probably be done soon.

lens0021:
I don't know why Github complains saying 'These conflicts are too complex to resolve in the web editor' but both `git rebase` and `git merge` do not show me any conflict.
Anyway, I will push the merged branch.

lens0021:
OK, I will fix the broken tests later.

Universal-Omega:
> I don't know why Github complains saying 'These conflicts are too complex to resolve in the web editor' but both `git rebase` and `git merge` do not show me any conflict. Anyway, I will push the merged branch.

Because I edited a file that was renamed.

Universal-Omega:
> OK, I will fix the broken tests later.

Thank you!

tests/phpunit/WikiManagerTest.php
```diff
@@ -1,5 +1,10 @@
 <?php
```
Universal-Omega, line 1: I believe that `@coversDefaultClass` needs to be the full namespaced class name.

Universal-Omega (changes requested):

Universal-Omega:
Also, the tests shouldn't have failed with "class 'WikiManager' not found" as they did. This means that `class_alias` is not properly working it seems. This is likely because the classes that are used for it need to be also added to AutoloadClasses it seems to load the alias class names as well. From what I can see, it seems that both the namespaced class and unnamespaced class need added to AutoloadClasses, even though it is in AutoloadNamespaces also. It does however seem that only RemoteWiki, WikiManager, and CreateWikiJson are actually used in other repositories, so potentially only them actually need to have the alias?

Universal-Omega:
Also, it seems `DBListGenerator` still needs namespaced.

I'm now considering this, but do we really want to convert to the src directory here, I did it for RemovePII following this PR because it would maintain consistency, but I'm not sure I like that due to the renaming causing loss of file history. I'd we really want to do this, and it has benefits for code standards, I'm willing, but if it's not worth it, maybe we should stick with includes?

lens0021:
> It does however seem that only RemoteWiki, WikiManager, and CreateWikiJson are actually used in other repositories, so potentially only them actually need to have the alias?

I hope to do it, but removing classes seems to be so unsafe for me because I am not sure I can search all code well, and/or there is no guarantee no one outside Miraheze depends on Miraheze extensions.

> Also, it seems DBListGenerator still needs namespace.

Thank you!

> I'm now considering this, but do we really want to convert to the src directory here, I did it for RemovePII following this PR because it would maintain consistency, but I'm not sure I like that due to the renaming causing loss of file history. I'd we really want to do this, and it has benefits for code standards, I'm willing, but if it's not worth it, maybe we should stick with includes?

I guess the file history is still there. The blame button on Github works fine and, in the way [this post](https://stackoverflow.com/a/2314757/10916512) on stack overflow describes, `git log --follow <path>` also shows me the history well. However, I am also not sure converting to the `src` directory is mandatory and the best practice, so if you agree, I would revert the converting.

Universal-Omega:
If possible I'd like to maintain includes then if using src doesn't provide any known benefit.

lens0021:
I've referenced https://gerrit.wikimedia.org/r/plugins/gitiles/mediawiki/extensions/GuidedTour/+/bc10db750db01ac18c31a919df8446f206b3a9f2%5E%21/#F0 but honestly, I don't understand why the long AutoloadClasses clauses are needed.

tests/phpunit/WikiManagerTest.php
```diff
@@ -35,8 +40,13 @@ public function testCreate() {
 
 		$wikiManager = new WikiManager( 'createwikitest' );
 
+		$this->assertInstanceOf( WikiManager::class, $wikiManager );
 		$this->assertNull( $wikiManager->create( 'TestWiki', 'en', 0, 'uncategorised', $user->getName(), $user->getName(), 'Test' ) );
 		$this->assertTrue( self::wikiExists( 'createwikitest' ) );
+
+		// Test class alias
+		$wikiManager = new \WikiManager( 'createwikitest' );
+		$this->assertInstanceOf( WikiManager::class, $wikiManager );
```
lens0021, line 49: I've added these lines to check whether the `class_alias` works.

tests/phpunit/WikiManagerTest.php
```diff
@@ -35,8 +40,13 @@ public function testCreate() {
 
 		$wikiManager = new WikiManager( 'createwikitest' );
 
+		$this->assertInstanceOf( WikiManager::class, $wikiManager );
```
Universal-Omega, line 43: What's the reasoning behind this test? There shouldn't be any reason at all this wouldn't match is there?

Universal-Omega:
> I hope to do it, but removing classes seems to be so unsafe for me because I am not sure I can search all code well, and there is no guarantee no one outside Miraheze depends on Miraheze extensions.

We are only concerned with the extensions in the Miraheze organisation here, as they'd have to manually update it if used externally, and they could then update any other extensions for it also. There should be no non Miraheze-maintained extensions depending on Miraheze-maintained ones. Also I did do an organisation-wide search on many of the classes from this PR, and only the 3 classes mentioned seem to have been used, though it's possible I missed some I don't think so as I don't personally recall other usages either, and I do think I'd remember if there were.

Universal-Omega:
I apologise for being so picky with this PR also, but it's quite a large change so just want to make sure we get it right, so we don't have to make so many changes later.

tests/phpunit/WikiManagerTest.php
lens0021: I intended to test the constructor itself, like some other WMF's extensions.[1] But I had to introduce WikiManagerTest::testConstruct() instead inserting code to existing WikiManagerTest::testCreate().
If your question is why constructors should be tested, I've thought and searched about it now, I haven't found the answer. Maybe because some calls to `new` make an exception raise? If you don't think these are needed, I will remove these in the next commit.

1: https://gerrit.wikimedia.org/g/mediawiki/extensions/AbuseFilter/+/bb3232ca72796586dc26b49618ebb7f5f6b6ca31/tests/phpunit/unit/CentralDBManagerTest.php#22

lens0021: To minimize the conflict, I've removed the addition to the test class as I think `class_alias` works fine for the last test.

Universal-Omega: Thanks!

Universal-Omega (approved):

PR #284, opened 2022-01-28
doc: Add why suppress_issue_types is specified

lens0021:
The original words from https://github.com/miraheze/CreateWiki/pull/283#discussion_r794250715

### DIFF
--- a/.phan/config.php
+++ b/.phan/config.php
@@ -14,7 +14,8 @@ $cfg['exclude_analysis_directory_list'] = array_merge(
 	]
 );
 
-// We intentionally explicitly specified this so that it picks up on deprecations, etc..
+// We explicitly specify this so that it
+// picks up on deprecations, and other issues
 $cfg['suppress_issue_types'] = [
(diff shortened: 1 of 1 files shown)


.phan/config.php
```diff
@@ -14,6 +14,7 @@
 	]
 );
 
+// We intentionally explicitly specified this so that it picks up on deprecations, etc..
```
Universal-Omega, line 17: Maybe the wording could be changed to this for better usage in this context?
> // We explicitly specify this so that it
> // picks up on deprecations, and other issues

Universal-Omega (changes requested):

.phan/config.php
lens0021: Thank you, I worried the syntax of the sentence I mixed is valid or not.

```suggestion
// We explicitly specify this so that it
// picks up on deprecations, and other issues
```

Universal-Omega (approved):

PR #287, opened 2022-01-28
set defaults for cw_wikis

Universal-Omega:


RhinosF1:
Schema change patch needs to be added

Universal-Omega:
> Schema change patch needs to be added

It wasn't really possible for wikis to even be created without setting the default values first, so I assumed it didn't need one, as CW is not used much externally and Miraheze must already since it needed to manually be set to a default when I did this for Betaheze.

RhinosF1:
I know it doesn't work not setting defaults but the schema on db must match what's on the repo.

In an amazing future world, we'll get a drift tracker.

Universal-Omega:
> I know it doesn't work not setting defaults but the schema on db must match what's on the repo.
> 
> In an amazing future world, we'll get a drift tracker.

#295 opened but I'm not able to do anything else with that right now.

PR #312, opened 2022-03-25
Create new CreateWikiRegexConstraint class

Universal-Omega:
### Benefits: ###
* Add regex validation and sanitisation
* Avoid failing the entire regex if one line or array key is invalid, rather just skip it and log in debug log that something went wrong
* Improve on-wiki comment disallowed words message regex — don't pull from `Title::newFromText()`, but rather `wfMessage()` directly
* Centralised regex functions

Also deprecates `MediaWiki:CreateWiki-blacklist`, replacing it with `MediaWiki:CreateWiki-disallowlist`, temporally maintaining current as fallback to allow for time for transition.

### DIFF
--- a/includes/CreateWikiRegexConstraint.php
+++ b/includes/CreateWikiRegexConstraint.php
@@ -101,9 +101,9 @@ class CreateWikiRegexConstraint {
 			if ( self::validateRegexes( [ $regex ] ) ) {
 				return $regex;
 			}
-		}
 
-		wfWarn( 'Contains invalid regex.' );
+			wfWarn( 'Contains invalid regex.' );
+		}
 
 		return '';
 	}
--- a/includes/CreateWikiRegexConstraint.php
+++ b/includes/CreateWikiRegexConstraint.php
@@ -21,6 +21,19 @@ class CreateWikiRegexConstraint {
 	}
 
 	/**
+	 * @param array &$regexes
+	 */
+	private static function filterInvalidRegexes( &$regexes ) {
+		foreach ( $regexes as $regex ) {
+			if ( !StringUtils::isValidPCRERegex( $regex ) ) {
+				wfWarn( 'Contains invalid regex.' );
+
+				unset( $regexes[$regex] );
+			}
+		}
+	}
+
+	/**
 	 * @param array $lines
 	 * @return array
 	 */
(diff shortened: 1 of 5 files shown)


includes/CreateWikiRegexConstraint.php
```diff
@@ -0,0 +1,122 @@
...
+	 * @param array &$regexes
+	 */
+	private static function filterInvalidRegexes( &$regexes ) {
+		foreach ( $regexes as $regex ) {
+			if ( !StringUtils::isValidPCRERegex( $regex ) ) {
+				wfWarn( 'Contains invalid regex.' );
+
+				unset( $regexes[$regex] );
+			}
+		}
```
lens0021, line 33: I belive the key of the element should be used to unset it:
```suggestion
		foreach ( $regexes as $key => $regex ) {
			if ( !StringUtils::isValidPCRERegex( $regex ) ) {
				wfWarn( 'Contains invalid regex.' );

				unset( $regexes[$key] );
			}
		}
```

includes/CreateWikiRegexConstraint.php
```diff
@@ -0,0 +1,122 @@
...
+	}
+
+	/**
```
lens0021, line 36: Though I can guess this function will remove comments and empty lines, It will be helpful if there is documentation. But please feel free to add or not.

includes/CreateWikiRegexConstraint.php
```diff
@@ -0,0 +1,122 @@
...
+			)
+		);
+	}
+
+	/**
+	 * @param string $text
+	 * @return array
+	 */
+	private static function regexesFromText( $text ) {
+		$lines = explode( "\n", $text );
```
lens0021, line 53:
```suggestion
		$lines = explode( '\n', $text );
```

includes/CreateWikiRegexConstraint.php
```diff
@@ -0,0 +1,122 @@
...
+			wfWarn( 'Contains invalid regex.' );
+		}
+
+		return $regexes;
+	}
+
+	/**
+	 * @param array $regexes
+	 * @param string $start
+	 * @param string $end
```
lens0021, line 66: Can these descriptions be improved? I think It is not easy to understand the context of the 'start' and 'end' by just reading their names.

lens0021 (changes requested):

includes/CreateWikiRegexConstraint.php
Universal-Omega: Done. Thanks!

includes/CreateWikiRegexConstraint.php
Universal-Omega: Done. Thanks!

includes/CreateWikiRegexConstraint.php
Universal-Omega: Done. Thanks!

lens0021:
The multiple `wfWarn()`s are confusing me. Have you considered throwing and catching an exception?

Universal-Omega:
> The multiple `wfWarn()`s are confusing me. Have you considered throwing and catching an exception?

We don't want to throw exceptions, but rather just warn that regex is invalid, and still process valid portions. It should however potentially use `wfDebugLog()` instead.

includes/CreateWikiRegexConstraint.php
```diff
@@ -0,0 +1,136 @@
...
+			]
+		);
+	}
+
+	/**
+	 * @param array &$regexes
+	 * @param string $name name of regex caller (config or message) for logging
+	 * @param string $start
+	 * @param string $end
+	 */
```
lens0021, line 28: I think indicating this function has no return value is helpful to inform the difference to the usage of `array_filter()`.
```suggestion
	 * @return void
	 */
```

Universal-Omega: Done. Thanks!

dmehus:
LGTM. :+1:

Thanks to @Universal-Omega and @lens0021 :)

PR #314, opened 2022-04-08
Use wdio-mediawiki/wdio-defaults.conf.js and run tests concurrently

lens0021:
[T226869 Run browser tests in parallel](https://phabricator.wikimedia.org/T226869)


lens0021:
Apologies, I was confused about the scope of wdio.conf.js.

PR #310, opened 2022-03-02
Update hook calling to use new HookContainer system

lens0021:
https://phabricator.wikimedia.org/T254670

PR #315, opened 2022-04-08
Update hook calling to use new HookContainer system

lens0021:
~~Split from https://github.com/miraheze/CreateWiki/pull/310 and follows it.~~

https://phabricator.wikimedia.org/T254670

Uses DI for HookRunner. To provide the HookRunner to WikiManager, RemoteWiki, etc, I added $hookRunner to their constructor.
~~This is not compatible with ManageWiki. But I believe CreateWiki can have a breaking change in some way I'v not known.~~


RhinosF1:
It's technically impossible to deploy a breaking change without causing an exception at some point so I'd rather the PR was wrote in a way that's possible to roll out without errors.

Universal-Omega:
> It's technically impossible to deploy a breaking change without causing an exception at some point so I'd rather the PR was wrote in a way that's possible to roll out without errors.

I can deploy it later with only 1-2 seconds of errors if ManageWiki is also fixed for support. So this is fine with me.

lens0021:
It sounds very fragile, I will change the new parameters as optional.

Universal-Omega:
Is this PR to replace #310? Asking since it also includes the same stuff as that does.

lens0021:
It can. Should I? I separated PRs by their purposes, for readability.

Universal-Omega:
> It can. Should I? I separated PRs by their purposes, for readability.

Well if the other is merged that would cause a major conflict, so probably easier to do all at once.

lens0021:
If the list of parameters of the constructors is being longer for other services like $hookRunner, they can be replaced with an instance of a new Context class, in the future.

PR #326, opened 2022-07-07
Notifications: use primary link instead of secondary links

Universal-Omega:
Per `PHP Notice: Trying to access array offset on value of type bool`:
https://github.com/wikimedia/mediawiki-extensions-Echo/blob/0a7c3d0/includes/formatters/EchoPlainTextEmailFormatter.php#L20, which is required, whereas https://github.com/wikimedia/mediawiki-extensions-Echo/blob/0a7c3d0/includes/formatters/EchoPlainTextEmailFormatter.php#L24 does the same thing with multiple, but is by default, an empty array (https://github.com/wikimedia/mediawiki-extensions-Echo/blob/0a7c3d0/includes/formatters/EventPresentationModel.php#L522), so does nothing.

### DIFF
--- a/includes/Notifications/EchoCreateWikiPresentationModel.php
+++ b/includes/Notifications/EchoCreateWikiPresentationModel.php
@@ -24,16 +24,10 @@ class EchoCreateWikiPresentationModel extends EchoEventPresentationModel {
 	}
 
 	public function getPrimaryLink() {
-		return false;
-	}
-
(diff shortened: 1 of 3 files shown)


Universal-Omega:
I actually opened https://gerrit.wikimedia.org/r/c/mediawiki/extensions/Echo/+/812365 instead, since I realised it was intended to be able to return false.

PR #336, opened 2022-09-01
add onhold to SpecialRequestWikiQueue status dropdown

The-Voidwalker:


The-Voidwalker:
Somewhat unrelated, but would it make sense to have these values have i18n?

Universal-Omega:
> Somewhat unrelated, but would it make sense to have these values have i18n?

Yes, that is something I have planned for a clean up in RequestWiki as a whole.

PR #337, opened 2022-09-14
Make setting a request on hold generate a log action

AgentIsai:
Title says it all


Universal-Omega:
This theoretically will work. Need to test.

AgentIsai:
I've tested it locally on my farm so from my experience, it works

Universal-Omega:
> I've tested it locally on my farm so from my experience, it works

I understand that, I usually just like testing things myself for confidence (and sanity) when merging. Not that I don't trust you whatsoever, just like to be certain myself before merging stuff for my own peace of mind.

PR #348, opened 2022-10-30
Sync i18n with Meta

AgentIsai:
CreateWiki's official system messages don't align with Meta's. On Meta, wiki requests are not 'accepted' but rather 'approved'. Equally, wiki requests are not 'placed on hold' but rather 'held.' No one has contested these changes so far so it may be beneficial to sync them officially to the actual extension so that new translations can reflect this.

Also syncs `requestwiki-header` to what Meta says so those in other languages can read the updated header.

And also adds support for log filtering of 'on hold' requests.


Universal-Omega:
In a couple months I plan to change the current version of things like `requestwiki-header` to something non-Miraheze specific and move those to MirahezeMagic, as these should ideally not be Miraheze specific here anyway. But this can probably be fine for now.

AgentIsai:
This is technically moot so closing.

PR #350, opened 2022-10-31
Redirect anonymous users to login

AgentIsai:
This PR redirects anonymous users to login rather than displaying the "You need to login first to request wikis" message which should make requesting wikis just a tiny bit more easier


includes/RequestWiki/SpecialRequestWiki.php
```diff
@@ -31,17 +31,10 @@ public function execute( $par ) {
 		$request = $this->getRequest();
 		$out = $this->getOutput();
 
+		$this->requireLogin( 'requestwiki-notloggedin' );
 		$this->setParameter( $par );
 		$this->setHeaders();
 
-		if ( !$this->getUser()->isRegistered() ) {
-			$loginurl = SpecialPage::getTitleFor( 'Userlogin' )->getFullURL( [ 'returnto' => $this->getPageTitle()->getPrefixedText() ] );
```
Universal-Omega, line 38: Remove the reference to `use SpecialPage;` at the very top.

Universal-Omega (changes requested):

Universal-Omega (approved):

includes/Hooks.php
```diff
@@ -27,7 +27,10 @@ public function __construct( Config $config, CreateWikiHookRunner $hookRunner )
 		$this->config = $config;
 		$this->hookRunner = $hookRunner;
 	}
-
+	public static function onLoginFormValidErrorMessages( &$messages ) {
```
Universal-Omega, line 30: This needs to:

1) Implement interface
2) Have no return value (remove `return true;`)
3) Fix param type declaration for `&$messages` to be `array &$messages`
4) Add `/** @inheritDoc */` comments
5) Use hook handler (non-static hook)
6) Fix spacing between the functions
7) Move below the `onRegistration()` callback below

I can help with some of this if needed also

Universal-Omega (changes requested):

includes/Hooks.php
AgentIsai: Should be done

Universal-Omega (approved):

PR #363, opened 2023-01-17
Display read-only errors when database is in read-only

AgentIsai:
When the database is in read-only mode, all forms are still usable. Some of them will throw a "DBReadOnlyError" once you click submit which is counterintuitive as a user might think they're fine to request a wiki/do an action but then will receive an error once they submit a form. Now, they'll see a read-only error which will prevent them from doing any action

### DIFF
--- a/includes/CreateWiki/SpecialCreateWiki.php
+++ b/includes/CreateWiki/SpecialCreateWiki.php
@@ -26,6 +26,7 @@ class SpecialCreateWiki extends FormSpecialPage {
 	protected function getFormFields() {
 		$par = $this->par;
 		$request = $this->getRequest();
+		$this->checkReadOnly();
 
 		$formDescriptor = [
 			'dbname' => [
--- a/includes/RequestWiki/SpecialRequestWiki.php
+++ b/includes/RequestWiki/SpecialRequestWiki.php
@@ -33,9 +33,9 @@ class SpecialRequestWiki extends FormSpecialPage {
 		$this->requireLogin( 'requestwiki-notloggedin' );
 		$this->setParameter( $par );
 		$this->setHeaders();
-		$this->checkReadOnly();
 
 		$this->checkExecutePermissions( $this->getUser() );
+		$this->checkReadOnly();
 
 		if ( !$request->wasPosted() && $this->config->get( 'CreateWikiCustomDomainPage' ) ) {
 			$customdomainurl = Title::newFromText( $this->config->get( 'CreateWikiCustomDomainPage' ) )->getFullURL();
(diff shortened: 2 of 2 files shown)


includes/RequestWiki/SpecialRequestWiki.php
```diff
@@ -35,6 +35,7 @@ public function execute( $par ) {
 		$this->setHeaders();
 
 		$this->checkExecutePermissions( $this->getUser() );
+		$this->checkReadOnly();
```
Universal-Omega, line 38: should be checked before permissions I think.

AgentIsai: Changed

includes/RequestWiki/SpecialRequestWiki.php
```diff
@@ -34,6 +34,7 @@ public function execute( $par ) {
 		$this->setParameter( $par );
 		$this->setHeaders();
 
+		$this->checkReadOnly();
```
Universal-Omega, line 37: Wait, this isn't needed, checkExecutePermissions has checkReadOnly in it?

Universal-Omega (changes requested):

includes/RequestWiki/SpecialRequestWiki.php
Universal-Omega: Just define requiresWrite as true instead.

Universal-Omega: Which is default so nevermind

AgentIsai: So is this redundant then?

includes/CreateWiki/SpecialCreateWiki.php
```diff
@@ -26,6 +26,7 @@ public function __construct( CreateWikiHookRunner $hookRunner ) {
 	protected function getFormFields() {
 		$par = $this->par;
 		$request = $this->getRequest();
+		$this->checkReadOnly();
```
Universal-Omega, line 29: This is also not needed, it is done by the default execute method of FormSpecialPage that is not overridden here.

Universal-Omega (changes requested):

includes/RequestWiki/SpecialRequestWiki.php
Universal-Omega: Yeah.

AgentIsai: That's funny. It didn't initially error me out when I set my own local test instance to read-only but did after I set this and even if I unset it. I'm guessing it was a momentary delay in something. Anyhow, something which does 100% work for a fact even in read-only is RequestWikiQueue so I'll focus on that instead

Universal-Omega:
Please reopen if needed but I think this is redundant per my review above over a year ago

PR #365, opened 2023-01-18
Rename the `requestaccept` log to `requestapprove`

Universal-Omega:
We will probably need to update the database also:

```sql
UPDATE logging SET log_action = 'requestapprove' WHERE log_action = 'requestaccept';
UPDATE recentchanges SET rc_log_action = 'requestapprove' WHERE rc_log_action = 'requestaccept';
```


Naleksuh:
I was asked to comment here. Does this do anything to front facing end? Looks like it's just changing it on the backend.

I already changed a few instances of "accept" to "approve" to make it more consistent, but I don't care what the code says. It can say either.

Universal-Omega:
> I was asked to comment here. Does this do anything to front facing end? Looks like it's just changing it on the backend.
> 
> I already changed a few instances of "accept" to "approve" to make it more consistent, but I don't care what the code says. It can say either.

The actual log action name when searching or filtering in Special:Log is used by this. I didn't know the message had already been changed on Meta either.

Naleksuh:
One message I would like changed is "User placed request X on hold" to "User held request X" or similar. Ideally the request number should be the last thing in the message and be made consistent with the others. Right now it's weird. I did change that one but Reception123 deleted it

PR #366, opened 2023-01-19
Update

Justman10000:
- Sounds pretty strange in German 😅


Universal-Omega:
This updated English, which we won't be doing in this way. You can update German like this in translatewiki.net to be this context if you need to though.

PR #375, opened 2023-01-27
Revert "Replace deprecated `wfGetDB()`"

Universal-Omega:
Reverts miraheze/CreateWiki#356

If this isn't an injected load balancer, then wfGetDB() is necessary. Otherwise it causes a drift, leading to errors with wiki creations.


The-Voidwalker:
Tested not working, I have a different fix prepared

PR #376, opened 2023-01-28
use DeferredUpdates for shell scripts

The-Voidwalker:
this ensures that if the cwdb writer gets stuck into a transaction, that transaction will be finished before the scripts get called


The-Voidwalker:
As long as I copied my changes from test131 properly, this should work.

Universal-Omega (approved):

PR #377, opened 2023-01-28
delay recacheJson to DeferredUpdates

The-Voidwalker:
this will ensure that even if another process triggers CreateWikiJson::update at the same time (causing a race condition to updating the databases.json list), the changes to cw_wikis will be committed and available to all processes that might be affected by this race condition.


The-Voidwalker:
(CI lint fixed failed checks)

PR #382, opened 2023-01-31
Try to add some qunit tests

Universal-Omega:


Universal-Omega:
Might try again later. Doesn't work for now. And not worth waisting time trying.

PR #383, opened 2023-02-02
Add link from wiki request back to queue

AgentIsai:
This PR adds a link from wiki requests back to the wiki request queue, akin to how subpages have a link under them which links back to the main page. Useful for requesters and for WCs to quickly go back to the queue.

### DIFF
--- a/includes/RequestWiki/SpecialRequestWikiQueue.php
+++ b/includes/RequestWiki/SpecialRequestWikiQueue.php
@@ -25,7 +25,7 @@ class SpecialRequestWikiQueue extends SpecialPage {
 		} else {
 			$this->lookupRequest( $par );
 
-			$this->getOutput()->setPageTitle( $this->msg( 'createwiki-viewrequest' )->text() );
+			$this->getOutput()->setPageTitle( $this->msg( 'requestwikiqueue-request-view' )->text() );
 
 			$link = $this->getLinkRenderer()->makeKnownLink(
 				static::getSafeTitleFor( 'RequestWikiQueue' ),
--- a/includes/RequestWiki/SpecialRequestWikiQueue.php
+++ b/includes/RequestWiki/SpecialRequestWikiQueue.php
@@ -24,6 +24,14 @@ class SpecialRequestWikiQueue extends SpecialPage {
 			$this->doPagerStuff();
 		} else {
 			$this->lookupRequest( $par );
+
+			$link = $this->getLinkRenderer()->makeKnownLink(
+				static::getSafeTitleFor( 'RequestWikiQueue' ),
+				$this->msg( 'requestwikiqueue' )->text()
+			);
+
+			$this->getOutput()->addSubtitle( "&lt; $link" );
+
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


Universal-Omega:
There is a returnto function to add it I think. Could be wrong though.

AgentIsai:
> There is a returnto function to add it I think. Could be wrong though.

I'm not seeing anything mentioned in the docs.

Universal-Omega:
> > There is a returnto function to add it I think. Could be wrong though.
> 
> I'm not seeing anything mentioned in the docs.

`$this->getOutput()->addReturnTo()`

Universal-Omega (changes requested):
Per above

includes/RequestWiki/SpecialRequestWikiQueue.php
```diff
@@ -24,6 +24,16 @@ public function execute( $par ) {
 			$this->doPagerStuff();
 		} else {
 			$this->lookupRequest( $par );
+
+			$this->getOutput()->setPageTitle( $this->msg( 'requestwikiqueue-request-view' )->text() );
```
Universal-Omega, line 28: I'm confused what the purpose of this is?

AgentIsai: See comment for commit. If this is merged, both the title and subtitle will read "Wiki request queue" which is a lot of repetition.

Universal-Omega: > See comment for commit. If this is merged, both the title and subtitle will read "Wiki request queue" which is a lot of repetition.

No, subtitle will read return to wiki request queue with my suggested changes, which is how everything in core and other extensions do it also. I don't see a reason to make that change here.

AgentIsai: The proposed code adds a "Return to" link at the bottom of the form.

Universal-Omega: > The proposed code adds a "Return to" link at the bottom of the form.

Yes, if you use same placement.

Universal-Omega: Try placing above lookupRequest

AgentIsai: But why not use a subtitle link as Special:ChangeCredentials, Special:GlobalPreferences, and Special:Interwiki/add do? The placement of the return to text is awkward at the bottom.

Universal-Omega: Because I dislike how that is done. Its a mess and goes against my plans to clean it up to use core functions. I am highly opposed to doing it the current way. addReturnTo() is one method designed for this, so thats why I want to. But if you want to use a subtitle. Do it simply with $this->getOutput()->addBacklinkSubtitle( $this->getPageTitle() );

AgentIsai: Understandable, noted for next time.

Universal-Omega:
Thanks! If you confirm testing works as you desired, I'll merge now.

AgentIsai:
I've tested it in my local instance and it works.

Universal-Omega:
Awesome! Thanks! Will merge when CI passes.

PR #413, opened 2023-06-05
revert access change without prior discussion

Naleksuh:
Previous behavior was intentional and user admitted it themself at a prior time. This is not a "logic flaw" and exists to seperate the duties of Meta sysops and Stewards. A change like this should not be done with a user merging their own change. Using commits to enact access changes goes outside the bounds of SRE

### DIFF
--- a/includes/RequestWiki/RequestWikiRequestViewer.php
+++ b/includes/RequestWiki/RequestWikiRequestViewer.php
@@ -116,7 +116,7 @@ class RequestWikiRequestViewer {
 			];
 		}
 
-		if ( $permissionManager->userHasRight( $userR, 'createwiki' ) && !$userR->getBlock() || $userR->getId() == $request->requester->getId() ) {
+		if ( $permissionManager->userHasRight( $userR, 'createwiki' ) || $userR->getId() == $request->requester->getId() ) {
 			$formDescriptor += [
 				'comment' => [
 					'type' => 'textarea',
@@ -208,7 +208,7 @@ class RequestWikiRequestViewer {
 			];
 		}
 
-		if ( $permissionManager->userHasRight( $userR, 'createwiki' ) && !$userR->getBlock() ) {
+		if ( $permissionManager->userHasRight( $userR, 'createwiki' ) ) {
 			$visibilityOptions = [
 				0 => wfMessage( 'requestwikiqueue-request-label-visibility-all' )->text(),
 				1 => wfMessage( 'requestwikiqueue-request-label-visibility-hide' )->text(),


Naleksuh:
Also this would apply to partial blocks as well, another problem with this change

MacFan4000:
It is a logic flaw, since if Special:CreateWiki is unusable when blocked, you shouldn't be able to create wikis even with autocreate. So, I believe this was the intended behavior.

MacFan4000:
> Also this would apply to partial blocks as well, another problem with this change

That can be fixed in a followup commit / a bug report can be filed on phabricator.

Naleksuh:
The reason why the form is enabled is to seperate Meta sysops from Stewards. I don't know why the page is disabled, but the ability to create wikis is tied to it for this reason. If it wasn't, it would alllow Meta sysops to take global actions, giving free reign ove who can and cannot create wikis.

Agent has previously admitted that this was intended behavior, but declared it a "logic flaw" today while editing user groups and self-merging this commit. This should be reopened because it is not a flaw and a change like this needs discussion, such as on phabricator as you mentioned

Naleksuh:
@AgentIsai Please revert your change. I understand that you did this shortly after changing user groups and this was no doubt tied to that, but there is no need to override via SRE, or merge your own changes. This defeats having a PR system in the first place

AgentIsai:
This was never intended behavior, you can't request wikis or manually create them while blocked, so why can you create them via RequestWikiQueue while still blocked? This is not a change of access, this is fixing a bug.

Naleksuh:
You can't request wikis because that form is open to the general public. And you discussed this previously:

2023-05-09 00:09:13 | <MirahezeRelay> | <BWM#6978> @Orduin to confirm something, are  indefinitely blocked users on Meta permitted to continue to approve/deny  wiki requests?
2023-05-09 00:09:23 | <MirahezeRelay> | <BWM#6978> as a piece of clarification
2023-05-09 00:14:48 | <MirahezeRelay> | <Agent#3928, replying to BWM#6978> They are

I see no mention of this being a bug. This seems to be tied to your recent user group change

AgentIsai:
They're permitted in a technical sense due to the bug. I've discussed this bug at length with other SRE members years prior to that message. Meta sysops have a lot of global reach, they have the ability to set global interwiki prefixes among with a lot of other global things. I don't see why we're all of a sudden wanting to curtail the global reach they've always had as a block also affects users ability to vote on RfCs, make SN/CN requests, and much more.

Naleksuh:
@AgentIsai Because it is the status quo. And if you disagreed with it, it would have made more sense to bring it up somewhere instead of self-merging a change to it. As for the other global reaches they have, I was not aware Meta sysops can do that, but regardless it does not really make sense to use one existing system to unilaterally change the status quo on an entirely different system, especially without any consensus. Especially when there was not any currently ongoing problem with leaving it how it is.

Naleksuh:
And this discussion would make more sense on Phabricator. It is somewhat the result of the self-merged change. Please revert your self-merged change and bring this up for discussion on Phabricator somewhere, in a similar manner to how the interface-admin 2FA issue was resolved

AgentIsai:
Sysadmins can merge their own changes, nothing precludes it, it is just common procedure to make a PR to allow CI to run and detect any bugs + to allow you to easily revert changes with the Revert button on GitHub. All this change does is fix a bug, that's all.

Naleksuh:
But the whole point of this is that it is not a bug, and even if it was, if something had been permitted for years it should not be changed unilaterally by one person, a change like this should have had discussion because it is the status quo. There is no limit to the number of changes that can just be enacted without consensus by anyone with merge access, but doing that is stepping outside the bounds of SRE. That is why I said a task for this should have been created at the very least.

AgentIsai:
@Reception123 As Engineering Manager for the MediaWiki team, you are final arbitrator in this manner. I will heed to any comment or admonishment you say.

Reception123:
First of all, I do believe it is quite absurd that Special:CreateWiki is unusable when one is blocked but they can still create wikis with RequestWiki and so that clearly demonstrates that there is a break in logic somewhere.
Second of all, in terms of the separation argument while I can understand the logic behind it it also doesn't make sense for someone who has been blocked by a Meta administrator to be able to still continue to create wikis which requires trust. Meta administrators are also elected by the community and in reality it's pretty much the same community. If a block is seen as unfair I'm sure that Stewards would be able to intervene and get it resolved, as that's their function after all. 
Therefore, I don't see this very thin separation being able to justify allowing a user who is blocked on Meta (by elected users and presumably for repeatedly violating policies) to create wikis. How would this look to an average user? I don't think it would make sense to them.
Usually I would recommend that you opened a discussion on Meta if you disagreed but obviously that's not possible due to circumstances.

Naleksuh:
Fair enough. My goal here was mainly to seperate between Meta sysops and stewards and avoid unilateral self-enacting of changes. While it is not technically prohibited to merge your own pull requests, I think it is still a bad idea. Just because you can, does not mean you should.

I would be interested in how these can be seperated through other means, and what has gone into the current system (for example, why there is a setting to restrict viewing a wiki request to the "delete" permission). But I also was not aware that CreateWiki worked this way until Agent pointed it out

PR #427, opened 2023-07-30
RequestWiki use zero instead of null if values missing

The-Voidwalker:
If the category, private, purpose, and bio features are turned off, these values will be missing from the form data. As the DB does not accept NULL for these values, we must fall back to a zero value.


theresnotime (approved):
lgtm

RhinosF1 (approved):
Looks fine, haven't checked default against sql but sensible approach and assuming void has.

PR #438, opened 2023-09-08
Make status messages on Special:RequestWikiQueue translatable

Dringsim:
On https://meta.miraheze.org/w/index.php?title=Special:RequestWikiQueue&dir=prev&dbname=&requester=&status=%2A&uselang=qqx, some messages are untranslated:
image
image


i18n/en.json
```diff
@@ -77,9 +78,6 @@
 	"requestwikiqueue-handle": "Handle request",
 	"requestwikiqueue-inreview": "In review",
 	"requestwikiqueue-onhold": "On hold",
-	"requestwikiqueue-pager-status-approved": "approved",
```
redbluegreenhat, line 80: Messages removed from en.json also need to be removed from qqq.json

redbluegreenhat (changes requested):
Looks good, just a thing regarding removing i18n messages.

redbluegreenhat (approved):

PR #463, opened 2024-01-28
Support additional subdomains

Universal-Omega:


includes/WikiManager.php
```diff
@@ -129,11 +130,34 @@ public function create(
 			return $this->doCreateDatabase();
 		}
 
+		// This is probably some of the worst code I have ever written
```
redbluegreenhat, line 133: `parse_url` got your back. [RequestSSL uses it](https://github.com/miraheze/RequestSSL/blob/4e0c5f9c3e63106c8b33e3e89adfd205b00ccbb8/includes/Specials/SpecialRequestSSL.php#L284) when validating the requested custom domain, that code ensures the custom domain has the `https://` protocol. Just keep https://gist.github.com/redbluegreenhat/568a275d6a6ecec04db5d1d2bead0518 in mind.

Looks like you want to ensure the url has the `https://` protocol. parse_url will work as long as `$url` has any protocol, even if it is a relative one, for example `//meta.miraheze.org`.

PR #479, opened 2024-02-27
Use modern maintenance functions and improve manageInactiveWikis

Universal-Omega:


redbluegreenhat (approved):
LGTM. Not much to say about this one really. I did wonder "why bother rebuilding Recent Changes on the checkLastWikiActivity.php script?", but then I thought, better be safe than sorry there. Since we've recently had a report of recent changes not updating (https://issue-tracker.miraheze.org/T11598), I think this is a good idea.

PR #480, opened 2024-02-29
CI: Expect fatal error on certain conditions

AgentIsai:


Universal-Omega:
A fatal error should really never be 'expected' especially in WikiManager context which should throw strings, as its used in scripts, which shouldn't throw regular exceptions.

Universal-Omega:
Instead it should use wfDebugLog

Universal-Omega:
or catch the string in the script and throw a script friendly $this->fatalError

Universal-Omega:
#482

PR #481, opened 2024-02-29
Revert "Return FatalError if trying to delet a wiki that does not exist"

Universal-Omega:
This reverts commit 8e0333a.


Universal-Omega:
This should not be how this is done, at least not *yet*. Something like https://github.com/miraheze/CreateWiki/pull/480#issuecomment-1970383075 would be better *for now*.

PR #487, opened 2024-03-22
T10683: Expose the canned responses to JavaScript

redbluegreenhat:
Exposes the canned responses as a variable on subpages of Special:RequestWikiQueue.

https://issue-tracker.miraheze.org/T10683

### DIFF
--- a/extension.json
+++ b/extension.json
@@ -103,10 +103,10 @@
 		"LoginFormValidErrorMessages": {
 			"handler": "default"
 		},
-		"ParserGetVariableValueSwitch": {
+		"MakeGlobalVariablesScript": {
 			"handler": "default"
 		},
-		"ResourceLoaderGetConfigVars": {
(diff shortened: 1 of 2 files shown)


redbluegreenhat:
Now that I think about it, that var can be pretty big. Do we want to load this everywhere? I'm thinking there might be some potential use case that would benefit from the canned responses being available wiki-wide, but maybe I should just restrict this to Special:RequestWikiQueue somehow...

redbluegreenhat:
For now, I'll restrict them to Special:RequestWikiQueue and only if the current user has the (createwiki) perm, if a use case appears that benefits from this being available everywhere I'll consider it then.

redbluegreenhat:
I'll restrict it to Special:RequestWikiQueue using the [MakeGlobalVariablesScript](https://www.mediawiki.org/wiki/Manual:Hooks/MakeGlobalVariablesScript) hook instead

PR #503, opened 2024-04-04
T12010: Introduce CreateWiki-specific suppression rights and suppression log

redbluegreenhat:
- Introduces a log where usage of the suppression feature can be audited. Visibility of this log is restricted to users with the `(createwiki-suppressionlog)` feature.
- Introduces CreateWiki-specific suppression rights, `(createwiki-deleterequest)` and `(createwiki-suppressrequest)`.

Only two "levels" of suppression exists now. It is recommended for sysadmins to manually patch their database so that wiki requests where `cw_visibility = 3` are set to `2`. Wiki requests with cw_visibility set to 3 are not exposed, but rather are invisible for even those with `(createwiki-suppressrequest)` until that is done.

https://issue-tracker.miraheze.org/T12010


redbluegreenhat:
Hmmm, this seems like a good chance to move suppression functionality to the `WikiRequest` class. Else I'll have to duplicate code down the line when I get around to continuing the REST API.

PR #507, opened 2024-04-10
Raidarr New Main Page Revamp

pixDeVl:
Updated i18n/en.json new main page content with https://meta.miraheze.org/wiki/Special:PermanentLink/374508, plus some copy edits.


Universal-Omega (changes requested):
This should not be merged yet. I will remove this review when it can be. I appreciate this PR though. I am debating in CW or use a main page revamp to move this out of CreateWiki directly. What stopped me last time was i18n would need redone from all languages, but if it would need done anyway, then it may be a good time to do. But I am undecided still.

OAuthority:
(Just a note that I am planning to finish the default loadout this weekend, which will make the i18n keys redundant)

pixDeVl:
> (Just a note that I am planning to finish the default loadout this weekend, which will make the i18n keys redundant) 

If that's the case, I'm assuming it would be best to just close this PR since I believe the swift based loadout don't work with Pull Requests?

PR #514, opened 2024-05-02
Remove colon from the database name header

colleirose:
It doesn't really need to be there, and none of the other ones have the colon:
image


Universal-Omega:
This is used other places as well where this now makes it inconsistent

redbluegreenhat:
Inconsistent is kind of the name of the game when it comes to CW tho :p

PR #516, opened 2024-05-02
Add an API endpoint to post comments to wiki requests

redbluegreenhat:


redbluegreenhat:
The `WikiRequest` class has to be changed to take instances of classes implementing the `UserIdentity` interface rather than instances of `User` before this can be merged.

redbluegreenhat:
Just so I don't forget, one of the things blocking this is that I want to add an utility class for handling wiki request suppression, which at the very minimum should allow you to retrieve `$visibilityConds`, which should help reduce code duplication a bit on the REST handlers.

Universal-Omega:
Just noting mostly to myself, after #567 the way this is done will be changed.

* Type hint all properties and methods
* Use `WikiRequestManager` service injection
* Use `WikiRequestManager->exists` rather than a try-catch
* Use `WikiRequestManager::VISIBILITY_CONDS` constant rather than current code duplication
* Use modern namespaces IE `MediaWiki\Config\Config`

Universal-Omega:
This is now being applied in #567

PR #543, opened 2024-07-18
Use ObjectCacheFactory

Universal-Omega:

PR #550, opened 2024-09-18
Use ObjectCacheFactory

AgentIsai:
Fixes CI


Universal-Omega:
* #543 but that breaks 1.42 till 1.43 need to fix it still lol

AgentIsai:
You're right. I don't know why Phan complains then...

PR #558, opened 2024-09-21
Add support for PHP cache file for extension-list

Universal-Omega:


paladox:
Should use a temp file because writing to a file being read will have race conditions like we experienced already.

Universal-Omega:
> Should use a temp file because writing to a file being read will have race conditions like we experienced already.

This should never be an issue on something that can only manually be ran or that is only created when it doesn't exist at all.

PR #577, opened 2024-10-11
Remove unneeded comments where it only says param or return types

Universal-Omega:
Type hints tell you this so the docs are unnecessary if they don't explain anything else.

### DIFF
--- a/includes/CreateWiki/CreateWikiLogFormatter.php
+++ b/includes/CreateWiki/CreateWikiLogFormatter.php
@@ -12,10 +12,6 @@ class CreateWikiLogFormatter extends LogFormatter {
 
 	private LinkRenderer $linkRenderer;
 
-	/**
-	 * @param LogEntry $entry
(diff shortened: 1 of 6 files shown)


RhinosF1 (changes requested):
These aren't unneeded. They are used for doc generation and it's mediawiki standard to include them.

Universal-Omega:
> These aren't unneeded. They are used for doc generation and it's mediawiki standard to include them.

Not anymore it seems.
https://phabricator.wikimedia.org/T258925

and https://phabricator.wikimedia.org/T343443 for example

RhinosF1:
That surprises me, the coding style documentation makes no reference to it and I wonder how much it's followed.

RhinosF1:
We should on the same notes probably actually attempt to generate docs and publish them somewhere for our maintained extensions.

Universal-Omega:
Hmm maybe we should keep these documentation and instead expensively document everything with more explanation which would also solve the redundancy.

PR #579, opened 2024-10-12
Add hook to pass certain extra field data from cw_requests to cw_wikis

Universal-Omega:
In the future, this will make us be able to remove the `addNewRow()` method from RemoteWikiFactory and be able to migrate things we use in mw-config or MirahezeMagic like `primary-domain` and `mediawiki-version` from our own custom table fields we added to the `cw_wikis` table to be natively possible in CreateWiki using the hook and the natively-available `wiki_extra` field.

It will also allow us to remove the fake variable in WikiDiscover, `$wgWikiDiscoverDescription` to actually make it not have to just read fake variable from ManageWikSettings and instead be possible by just pulling the description from this `wiki_extra` field.

The other benefit this will bring will also be allowing to select the primary domain (which we use a custom field and hook in ManageWiki for) directly from RequestWiki without having to rely on our custom fields or an additional custom field in `cw_requests`, so this will allow us to bring selecting the primary domain directly from RequestWiki as well.

This has many more benefits for external use as well, making CreateWiki much more expandable utilizing hooks rather than custom very hacky overrides.


includes/Services/RemoteWikiFactory.php
```diff
@@ -369,6 +372,28 @@ public function unMarkExperimental(): void {
 		$this->newRows['wiki_experimental'] = 0;
 	}
 
+	public function getExtraFieldData( string $field ): mixed {
+		return $this->extra[$field] ?? null;
+	}
+
+	public function setExtraFieldData( string $field, mixed $value ): void {
+		if ( $value !== $this->getExtraFieldData( $field ) ) {
```
BlankEclair, line 380: [Early return](https://gomakethings.com/the-early-return-pattern-in-javascript/)?

Universal-Omega: Typically I like early returns but decided not to here as my is to eventually make RemoteWikiFactory more similar to how WikiRequestManager works which means it being more inline with that here makes it a tiny bit easier down the line which is why I decided against it. But I guess it wouldn't be horrible to early return also...

includes/Services/RemoteWikiFactory.php
```diff
@@ -369,6 +372,28 @@ public function unMarkExperimental(): void {
...
+			$newExtra = json_encode( $extra );
+
+			if ( $newExtra === false ) {
+				// Can not set invalid JSON data to wiki_extra.
+				return;
+			}
+
+			$this->extra = $extra;
+			$this->trackChange( $field, $this->getExtraFieldData( $field ), $value );
+			$this->newRows['wiki_extra'] = $newExtra;
```
Universal-Omega, line 393: **Personal Note**: This definitely needs tested with more than one field at once (calling this function more than once) as unlike RequestWiki we can't have just a set all field data at once function here, so like the first version of WikiRequestManager this may have an issue with that and will need to be tested and handled some other way if so.

includes/Services/RemoteWikiFactory.php
BlankEclair: Alright then, feel free to resolve this conversation if you decide not to early return

PR #598, opened 2024-10-17
Test

Universal-Omega:


Universal-Omega:
@coderabbitai Create me PHPUnit tests with 100% coverage for maintenance/manageInactiveWikisV2.php using MediaWiki's MaintenanceBaseTestCase. Use dataProvider when necessary, also use Generators if in DataProvider, and type hint all returns and params (even if return is void). Also include a coversDefaultClass comment and covers ::method. The test namespace should be in Miraheze\CreateWiki\Tests\Maintenance. Remember for MaintenanceBaseTestCase we need to override getMaintenanceClass and have it return ManageInactiveWikisV2::class so that $this->maintenance can be used as well.

PR #605, opened 2024-10-22
Implement fast track deletion for unused wikis

pixDeVl:
This was requested and approved by the community in https://meta.miraheze.org/wiki/Requests_for_Comment/Content_and_Dormancy_Policies back in March(over half a year ago, so long overdue and then some).
Since many, many, many wiki requests unfortunately never really get into editing and just give up on or forget about making entirely, leaving swarms of ghost wikis on the site taking up space.
This pull request implements the 'fast-track' deletion, where wikis that have no edits(besides MediaWiki default) are marked as inactive after 15 day, and deleted after 30. 
Rest in peace our 10k badge. 

## Summary by CodeRabbit

- **Refactor**
	- Restructured configuration settings to distinguish between standard activity and no-edit scenarios, resulting in refined thresholds for inactivity, closure, removal, and deletion.
	- Adjusted deletion grace period logic to enforce consistent behavior based on the updated configurations.
	- Improved activity log filtering by excluding default system entries.
	- Strengthened error handling for missing configurations, promoting smoother operation overall.


extension.json
```diff
@@ -369,11 +369,18 @@
...
+					"closed": 30,
+					"removed": 0,
+					"deleted": 0
+				},
+				"default": {
+					"inactive": 45,
+					"closed": 15,
+					"removed": 120,
+					"deleted": 7
+			}}
```
BlankEclair, line 383: Mangled JSON

maintenance/checkLastWikiActivity.php
```diff
@@ -42,18 +42,20 @@ public function execute(): void {
 	}
 
 	private function getTimestamp(): int {
+		$default_id = $this->getServiceContainer->getUserFactory->newFromName("MediaWiki default")->getActorId()
```
BlankEclair, line 45: Missing `;` at the end of the line, and missing `()` for a function call to `$this->getServiceContainer()`

maintenance/checkLastWikiActivity.php
```diff
@@ -42,18 +42,20 @@ public function execute(): void {
...
 		$dbr = $this->getReplicaDB();
 		$timestamp = $dbr->newSelectQueryBuilder()
 			->select( 'MAX(rc_timestamp)' )
 			->from( 'recentchanges' )
 			->where( [
 				$dbr->expr( 'rc_log_type', '!=', 'renameuser' ),
 				$dbr->expr( 'rc_log_type', '!=', 'newusers' ),
+				$dbr->expr( 'rc_actor', '!=', $default_id)
```
BlankEclair, line 53: Missing space before `)` and missing trailing comma

maintenance/manageInactiveWikisV2.php
```diff
@@ -83,6 +78,14 @@ private function checkLastActivity(
...
 		$lastActivityTimestamp = $activity->timestamp;
 
+		$track = ($lastActivityTimestamp != 0) ? 'default': 'no-edits'
+
+		$inactiveDays = (int)$this->getConfig()->get( ConfigNames::StateDays )[$track]['inactive'];
+		$closeDays = (int)$this->getConfig()->get( ConfigNames::StateDays )[$track]['closed'];
+		$removeDays = (int)$this->getConfig()->get( ConfigNames::StateDays )[$track]['removed'];
+		$canWrite = $this->hasOption( 'write' );
+
+
```
BlankEclair, line 88: Superfluous newline

maintenance/manageInactiveWikisV2.php
```diff
@@ -83,6 +78,14 @@ private function checkLastActivity(
 
 		$lastActivityTimestamp = $activity->timestamp;
 
+		$track = ($lastActivityTimestamp != 0) ? 'default': 'no-edits'
```
BlankEclair, line 81: Missing spaces around parenthesis and missing semicolon, and `!==` please?

BlankEclair (changes requested):

maintenance/checkLastWikiActivity.php
pixDeVl: Is a trailing comment on the last line of a list really needed? Regardless, added(these are on local, resolving to keep track before committing)

maintenance/manageInactiveWikisV2.php
pixDeVl: damn php and mediawiki standards..

BlankEclair (commented):
Apologies for the delay--was busy setting up my new phone, and you know how that is... I have to copy _everything_, then configure all my settings, then reverse engineer one of my banking apps to figure out how it detects root. Y'know, all those typical things

maintenance/checkLastWikiActivity.php
BlankEclair: Technically not needed, but it's MediaWiki convention I think (and definitely Python)

maintenance/manageInactiveWikis.php
```diff
@@ -51,7 +51,7 @@ public function execute(): void {
 		foreach ( $res as $row ) {
 			$dbName = $row->wiki_dbname;
 			$remoteWiki = $remoteWikiFactory->newInstance( $dbName );
-			$inactiveDays = (int)$this->getConfig()->get( ConfigNames::StateDays )['inactive'];
+			$inactiveDays = (int)$this->getConfig()->get( ConfigNames::StateDays )['default']['inactive'];
```
BlankEclair, line 54: I wonder if this should be no-edits?

pixDeVl: I think so? I'll do some thinking, but also need to look to make sure we're not breaking any previously implemented assumptions about the state from when there was only one track.

BlankEclair: If the `no_edits_inactivity < default_inactivity`, but we only do checking for when it's default, then the no edits stuff won't trigger, right?

pixDeVl: Updated to no-edits, but we should take a look to see if that doesn't break and assumptions and cause it to early close any wikis it shouldn't. A bit late for me to be doing a thoughtful look right now though.

BlankEclair:
Looks ~~gay~~ good to me

BlankEclair (approved):

maintenance/manageInactiveWikis.php
BlankEclair: cc @Universal-Omega

Universal-Omega (changes requested):
Mostly very minor nitpick stuff, but one critical issue as well.

maintenance/manageInactiveWikis.php
```diff
@@ -83,6 +79,13 @@ private function checkLastActivity(
 
 		$lastActivityTimestamp = $activity->getTimestamp();
 
+		$track = ( $lastActivityTimestamp !== 0 ) ? 'default': 'no-edits'
```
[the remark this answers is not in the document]
pixDeVl: @Universal-Omega  configured incorrectly being not defining no edits or default or just not defining no edits?

maintenance/ManageInactiveWikis.php
```diff
@@ -48,8 +48,9 @@ public function execute(): void {
 			->fetchFieldValues();
 
 		foreach ( $wikis as $wiki ) {
-			$remoteWiki = $remoteWikiFactory->newInstance( $wiki );
-			$inactiveDays = (int)$this->getConfig()->get( ConfigNames::StateDays )['inactive'];
+			$dbname = $wiki->wiki_dbname;
```
Universal-Omega, line 51: Remove variable and just use $wiki like before

maintenance/ManageInactiveWikis.php
```diff
@@ -48,8 +48,9 @@ public function execute(): void {
...
+			$remoteWiki = $remoteWikiFactory->newInstance( $dbname );
+			$inactiveDays = (int)$this->getConfig()->get( ConfigNames::StateDays )['no-edits']['inactive'];
```
Universal-Omega, line 53: I know it's already here but 1) why hardcover 'no-edits' and 2) move this out of the loop no need to recreate the variable every time.

Universal-Omega (changes requested):

maintenance/CheckLastWikiActivity.php
```diff
@@ -19,6 +19,10 @@ public function execute(): void {
 	}
 
 	public function getTimestamp(): int {
+		$defaultActor = $this->getServiceContainer()->getUserFactory()
```
Universal-Omega, line 22: Probably should have a test case in CheckLastWikiActivityTest to make sure properly excludes MediaWiki default edits.

maintenance/ManageInactiveWikis.php
Universal-Omega: I did this myself

maintenance/ManageInactiveWikis.php
Universal-Omega: I did this myself

maintenance/ManageInactiveWikis.php
```diff
@@ -32,6 +32,8 @@ public function execute(): void {
 			$this->fatalError( 'Enable $wgCreateWikiEnableManageInactiveWikis to run this script.' );
 		}
 
+		$inactiveDays = (int)$this->getConfig()->get( ConfigNames::StateDays )['no-edits']['inactive'];
```
Universal-Omega, line 35: Is there a reason we hardcode no-edits here? I'm uncertain about that.

maintenance/ManageInactiveWikis.php
Universal-Omega, line 1: This also needs tests in ManageInactiveWikisTest but I can probably put some work into doing that at some point.

Universal-Omega (changes requested):

Universal-Omega (changes requested):
Also needs updates in https://github.com/miraheze/CreateWiki/blob/master/includes/Services/WikiManagerFactory.php#L315 to support new format

maintenance/ManageInactiveWikis.php
pixDeVl: Performance wise, the check this is used in is to tell if it's worth doing more processing. Using the function to get the proper check would mean that code(querying the database) runs for every wiki even if they're active, while if we leave it here, [`checklastActivity()`](https://github.com/miraheze/CreateWiki/blob/b82de09d0e1115fb65c0bdee82b983e053e0c142/maintenance/ManageInactiveWikis.php#L67) would run on every wiki that hasn't been edited for at least 7 days under default config. So I guess it depends on which performance hit we think will hurt less? It would probably be activity->getTimestamp() I feel.. in which case we should change it.

pixDeVl:
@coderabbit review

includes/Services/WikiManagerFactory.php
```diff
@@ -311,7 +311,7 @@ public function delete( bool $force ): ?string {
 		$unixNow = (int)wfTimestamp( TS_UNIX, $this->dbw->timestamp() );
 
 		// Return error if the wiki is not deleted, force is not used, or the deletion grace period has not passed.
-		$deletionGracePeriod = (int)$this->options->get( ConfigNames::StateDays )['deleted'] * 86400;
+		$deletionGracePeriod = (int)$this->options->get( ConfigNames::StateDays )["no-edits"]['deleted'] * 86400;
```
BlankEclair, line 314:
```suggestion
		$deletionGracePeriod = (int)$this->options->get( ConfigNames::StateDays )['no-edits']['deleted'] * 86400;
```

maintenance/ManageInactiveWikis.php
```diff
@@ -68,18 +69,33 @@ private function checkLastActivity(
...
+
+		if ( !isset( $stateConfig['default'] ) ) {
+			throw new ConfigException(
+				'Default state config not found. Please check your configuration.'
+			);
+		}
+		if ( !isset( $stateConfig['no-edits'] ) ) {
+			$stateConfig['no-edits'] = $stateConfig['default'];
+			$this->output(
+				'No edits state config not found, using default instead.'
```
BlankEclair, line 89:
```suggestion
				"No edits state config not found, using default instead.\n"
```

maintenance/CheckLastWikiActivity.php
```diff
@@ -21,6 +21,10 @@ public function execute(): void {
...
+			->newFromName( 'MediaWiki default' )
+			->getActorId();
```
The-Voidwalker, line 26:
```suggestion
			?->getActorId();
```

`UserFactory->newFromName` can return null, so we can make this nullsafe. Not sure how well `$dbr->expr( 'log_actor', '!=', $defaultActor ),` would take a null value for `$defaultActor` though. Might be fine.

maintenance/ManageInactiveWikis.php
```diff
@@ -52,9 +53,17 @@ public function execute(): void {
 			$this->fatalError( 'Enable $wgCreateWikiEnableManageInactiveWikis to run this script.' );
 		}
 
+
+		$inactiveDays = (int)$this->getConfig()->get( ConfigNames::StateDays )['no-edits']['inactive'];
+
+		$databaseUtils = $this->getServiceContainer()->get( 'CreateWikiDatabaseUtils' );
+		$remoteWikiFactory = $this->getServiceContainer()->get( 'RemoteWikiFactory' );
+
+
```
The-Voidwalker, line 62:
```suggestion
```

These variables are no longer used here and can be removed.

maintenance/ManageInactiveWikis.php
```diff
@@ -52,9 +53,17 @@ public function execute(): void {
...
 		$this->initServices();
 		$dbr = $this->databaseUtils->getGlobalReplicaDB();
 
+
```
The-Voidwalker, line 66:
```suggestion
```

empty line cleanup!

maintenance/ManageInactiveWikis.php
```diff
@@ -66,6 +75,7 @@ public function execute(): void {
 			->fetchFieldValues();
 
 		foreach ( $wikis as $wiki ) {
+
```
The-Voidwalker, line 78:
```suggestion
```

Empty line cleanup!

maintenance/ManageInactiveWikis.php
```diff
@@ -66,6 +75,7 @@ public function execute(): void {
...
 			$remoteWiki = $this->remoteWikiFactory->newInstance( $wiki );
 			$inactiveDays = (int)$this->getConfig()->get( ConfigNames::StateDays )['inactive'];
```
The-Voidwalker, line 80:
```suggestion
			$inactiveDays = (int)$this->getConfig()->get( ConfigNames::StateDays )['no-edits']['inactive'];
```

Dunno if this is the correct state config we'd want to use here, but it needs to be changed.

The-Voidwalker (changes requested):

PR #606, opened 2024-10-22
Add support for request flagging

Universal-Omega:


includes/RequestWiki/FlaggedWikisPager.php
```diff
@@ -0,0 +1,151 @@
...
+
+				$formatted = '-';
+				break;
+			case 'cw_flag_reason':
+				$formatted = $this->escape( $row->cw_flag_reason );
+				break;
+			case 'cw_flag_actor':
+				$formatted = Linker::userLink(
+					$this->userFactory->newFromActorId( $row->cw_flag_actor )->getId(),
+					$this->userFactory->newFromActorId( $row->cw_flag_actor )->getName()
```
BlankEclair, line 84: Perhaps use an intermediary variable to avoid having to call `UserFactory::newFromActorId()` twice?

Justman100:
What would do this 👀?

BlankEclair:
https://issue-tracker.miraheze.org/T10527

PR #612, opened 2024-10-31
add NUMBEROFOPENREQUESTS magicword

Reception123:


BlankEclair (changes requested):
You need to add the new magic word to CreateWikiMagic.php, otherwise Bad Things Will Happen

PR #621, opened 2024-11-02
Migrate to use DatabaseVirtualDomains (T12376)

Universal-Omega:


BlankEclair (approved):
Haven't looked at the API semantics, but it looks good enough to me. If you want, I can take a closer look :p

PR #622, opened 2024-11-02
Don't fail if cluster is empty

Universal-Omega:
Also has a minor performance boost since it removes one query that we no longer need.


BlankEclair (approved):

BlankEclair:
Haven't checked API semantics but it LGTM

BlankEclair (approved):

PR #624, opened 2024-11-04
Add PopulateGlobalWiki maintenance script

Universal-Omega:


BlankEclair (approved):
god yes please

includes/Services/CreateWikiDataFactory.php
```diff
@@ -220,6 +221,13 @@ public function resetWikiData( bool $isNewChanges ): void {
 			->fetchRow();
 
 		if ( !$row ) {
+			if ( $this->wiki === $this->options->get( ConfigNames::GlobalWiki ) ) {
```
BlankEclair, line 224: Can this only trigger for maintenance scripts I wonder...

Universal-Omega: It can but I dont think it is worth it. If we don't have a global wiki we need to run a maintenance script, any extra behavior caused by this not being present can be accepted I think. It means the script needs ran. And it should automatically be ran if using updater. If you aren't then you'll just have to know to I guess. But if you really believe we should handle it, I can add more handling here.

BlankEclair: Eh, I don't really care

PR #636, opened 2024-11-12
Removing Approval Score thing

joelemiliano:
I hope I removed all AI stuff that makes that Approval Score thing pop up in RequestWiki comments because if not: CreateWiki blows up

But well, the reason is that... per Agent, that thing shouldn't be displayed


Universal-Omega:
No we don't want to remove this entirely.

Universal-Omega:
I appreciate the attempt here though but all that was actually needed was https://github.com/miraheze/mw-config/commit/7a3ea962cf6944810ae9005f22593c4f4838a11f

joelemiliano:
> I appreciate the attempt here though but all that was actually needed was [miraheze/mw-config@7a3ea96](https://github.com/miraheze/mw-config/commit/7a3ea962cf6944810ae9005f22593c4f4838a11f)

sorry I didn't know that :(

PR #656, opened 2025-02-13
This are wikis, not simply "sites"

Justman100:


Reception123:
We've used sitename for this variable ever since Miraheze's creation and there's really no need to change that.

PR #659, opened 2025-02-15
Use PHP8 property promotion

Universal-Omega:


BlankEclair (approved):
TIL we can do that, very handy :3

PR #671, opened 2025-02-21
eslint: enable and fix no-jquery/no-global-selector

Universal-Omega:


Universal-Omega:
> It feels kinda hacky to reformat the code in this way to appease the linter, but I suppose no linter can be perfect...

I don't actually know how to fix this, it was just a random attempt that didn't actually work. I'll try other things later maybe or just forget it...

PR #673, opened 2025-02-24
Apply SchemaVars when creating databases

Universal-Omega:


Universal-Omega:
Seems like this might actually already do this. So maybe this isn't needed actually... https://github.com/wikimedia/mediawiki/blob/3abc64a5d82067c0bc6b5d97efe4c1626fab6a2f/includes/libs/rdbms/database/Database.php#L2796

PR #681, opened 2025-02-28
Use strict_types

Universal-Omega:


BlankEclair (approved):
I skimmed through this, but how else do I review this? xD

PR #695, opened 2025-03-06
Cleanup/modernize maintenance scripts

Universal-Omega:
### Maintenance Scripts Now Require `run.php`

Maintenance scripts **can no longer be run directly** by passing them to the PHP interpreter. Instead, they now **require** `run.php`.

#### New Usage:
Scripts can now be executed using either of the following methods:

- **Without PHP explicitly:**
  ```sh
  maintenance/run CreateWiki:ListDatabases
  ```
- **Using PHP:**
  ```sh
  php maintenance/run.php CreateWiki:ListDatabases
  ```

This change significantly simplifies calling maintenance scripts, reducing the steps needed.

#### Impact on Miraheze:

For Miraheze, this means scripts can now be executed using mwscript more conveniently. For example:

```sh
mwscript CreateWiki:ListDatabases
```

### DIFF
--- a/extension.json
+++ b/extension.json
@@ -340,7 +340,7 @@
 			"value": false
 		},
 		"CreateWikiEnableManageInactiveWikis": {
-			"description": "Boolean. Whether the manageInactiveWikis.php maintenance script be enabled. That script marks wikis as inactive, closed, or deleted per $wgCreateWikiStateDays.",
+			"description": "Boolean. Whether the ManageInactiveWikis maintenance script be enabled. That script marks wikis as inactive, closed, or deleted per $wgCreateWikiStateDays.",
 			"value": false
 		},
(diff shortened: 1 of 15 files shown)


BlankEclair:
The WMF doesn't appear to do this yet? https://gerrit.wikimedia.org/r/plugins/gitiles/mediawiki/core/+/86a54423641cdf11bfb7274dc4af7558842211b9/maintenance/rebuildall.php#70

Code itself looks good enough to me, but I'm not sure if we need this :p

Universal-Omega:
Some extensions do it (but very few) and docs say to do it like this too.

Universal-Omega:
https://github.com/wikimedia/mediawiki/blob/86a54423641cdf11bfb7274dc4af7558842211b9/maintenance/README#L69

PR #702, opened 2025-03-08
RenameWiki: Check if the new database exists before moving

BlankEclair:
Doing so would've prevented T13319


Universal-Omega:
Per our conversation this won't actually work since the wiki is expected not to exist yet.

PR #705, opened 2025-03-09
DeleteWikis: rewrite to improve robustness and merge DeleteWiki into it

Universal-Omega:


maintenance/DeleteWikis.php
```diff
@@ -6,37 +6,115 @@
...
+		}
+	}
+
 	public function execute(): void {
+		$user = $this->getOption( 'user' );
+		if ( !$user ) {
+			$this->fatalError( 'Please specify the username of the user executing this script.' );
+		}
+
+		$this->deletedWikis = [];
```
BlankEclair, line 46: Should we set `$this->notified = false` here just in case?

Universal-Omega: Done

BlankEclair (approved):

PR #706, opened 2025-03-10
Add EmailBureaucrats maintenance script

Universal-Omega:


maintenance/EmailBureaucrats.php
```diff
@@ -0,0 +1,65 @@
...
+
+		$notificationData = [
+			'type' => 'custom-email',
+			'from-name' => $this->getOption( 'from-name' ),
+			'subject' => $this->getOption( 'subject' ),
+			'body' => [
+				'html' => $bodyHtml,
+				'text' => $bodyText,
+			],
+		];
```
BlankEclair, line 56: This should be handled such that only a text/plain email is sent without --parse-wikitext, and a multipart email with --parse-wikitext

Universal-Omega: Isn't that automatically handled by UserMailer?

BlankEclair: It looks like if --parse-wikitext is not passed, then the raw wikitext will be sent as the contents of both the text/plain and text/html bodies. We should pass in the body as the string if we're not doing HTML emails.

(And no, UserMailer does not deduplicate it if both are the same: https://gerrit.wikimedia.org/r/plugins/gitiles/mediawiki/core/+/8f3106b5702cc04a9877108f5fdb95b0947911b7/includes/mail/UserMailer.php#297)

Universal-Omega: Well if --parse-wikitext is not provided I just assumed it would just assume the user passes plain text without wikitext at which case both html and text would be the same thing no mattering if html email is enabled. Since it's an explicit argument if a user has wikitext in the body that we are passing id assume they'd pass with parse wikitext otherwise html just becomes text anyway.

BlankEclair: As a (hopefully) more helpful message,
```suggestion
		$body = $this->hasOption( 'parse-wikitext' )
			? [ 'text' => $bodyMessage->text(), 'html' => $bodyMessage->parse() ]
			: $bodyMessage->text();

		$notificationData = [
			'type' => 'custom-email',
			'from-name' => $this->getOption( 'from-name' ),
			'subject' => $this->getOption( 'subject' ),
			'body' => $body,
		];
```

Universal-Omega: I understand what you mean but in a maintenance script where you explicitly pass parameters if you pass --parse-wikitext or not the same output will be given because if you have it then it parses it if you dont it's plain text anyway. It seems like it'd just be making it more complicated if we handle it that way when a maintenance script means the user controls the input if --parse-wikitext is given wrong then that's the usage of the script that's wrong. I'm not entirely opposed to doing this, I'm not sure I see the absolute need to do it. But if you really feel strongly about it I guess we can do it like that.

Universal-Omega: My meaning is this basically, --parse-wikitext is basically just a flag telling the script it has wikitext. If it does not have wikitext then the html version of the email will be the same as the text version anyway because it has no wikitext to begin with. It acts under the assumption if the user provides wikitext the flag is parsed or to the script it is just plain text anyway.

BlankEclair: Email clients that can render HTML emails can also render plaintext emails too, albeit generally in a monospace font.

If we don't want the monospace font, then we should do something like `nl2br( $bodyMessage->escaped() )` for the HTML part

PR #777, opened 2025-09-14
Improve dark mode styling

lihaohong6:
Use codex design tokens to replace hard-coded colors for Special:RequestWikiQueue details.

Tested on Chrome/Firefox's dev tools.


lihaohong6:
Looks like CI wants me to change `border-color: var(--border-color-base, #c8ccd1);` into `border-color: var( --border-color-base, #c8ccd1 );`, which is pretty silly since [MDN doesn't do that](
https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_cascading_variables/Using_CSS_custom_properties). I do see some WMF repos such as vector add spaces like that, so maybe I'll submit to the almighty CI.

Universal-Omega (changes requested):
Okay so, while this may work, we should probably use LESS tokens. To do so, add
`@import 'mediawiki.skin.variables.less';` at the top then for example, instead of `var( --border-color-subtle, #ccc )` use `@border-color-subtle`.

PR #813, opened 2026-03-22
Add a hook for modifying the Special:CreateWiki form

lihaohong6:
For the CreateWikiLoadout extension. 

It's also nice to be able to add additional fields to Special:CreateWiki if similar needs arise in the future.

Only tested on localhost.


Universal-Omega:
Just FYI Special:CreateWiki will eventually be removed and replaced with a auto-create-via-request feature in Special:RequestWiki for better log management and lookup, and tracking creations better. It should hardly ever be used in production anyway. Either way no real issues with this though, but it doesnt do all that much good for CreateWiki since that will hardly ever be used at all.

PR #818, opened 2026-05-01
allow wiki requestors to abandon their request (T14115)

Reception123:


includes/Services/WikiRequestViewer.php
```diff
@@ -273,6 +273,19 @@ public function getFormDescriptor(): array {
 				'disabled' => $this->wikiRequestManager->isLocked(),
 				'section' => 'editing',
 			];
+
+			$isRequester = $user->getActorId() === $this->wikiRequestManager->getRequester()->getActorId();
+			if ( $isRequester && !$this->wikiRequestManager->isLocked() ) {
+				$abandonableStatuses = [ 'inreview', 'moredetails' ];
```
Universal-Omega: What about on hold, should we allow those to be abandonable also?

Universal-Omega: If not we should add that as a comment for future clarity as well.

Reception123: Oh yeah, I didn't think of that. I guess so yeah, if the user thinks it's not worth waiting for a potential second review and doesn't want to go through with it anymore.

includes/Services/WikiRequestViewer.php
```diff
@@ -273,6 +273,19 @@ public function getFormDescriptor(): array {
...
+				$canAbandon = in_array( $this->wikiRequestManager->getStatus(), $abandonableStatuses, true );
+				$formDescriptor['submit-abandon'] = [
+					'type' => 'submit',
+					'buttonlabel-message' => 'requestwiki-label-abandon-request',
+					'flags' => $canAbandon ? [ 'destructive' ] : [],
+					'disabled' => !$canAbandon,
+					'section' => 'editing',
+				];
+			}
```
Universal-Omega, line 286: We dont need to disable the `submit-abandon` button, we should only add it at all when abanonable to avoid confusing users. I would move the `$canAbandon` above the outer check, and add to that. Also the variable name `$abandonableStatuses` just reads odd to me. What I think is a better approach is adding a new constant in WikiRequestManager called `CAN_ABANDON_STATUSES` and then call `WikiRequestManager::CAN_ABANDON_STATUSES` in both places to ensure we dont accidentally change in one place but not others in the future as well.

Secondly, logic is a little complex and duplicated. I think we should add a new helper called `canAbandonRequest` in which as an argument to pass a UserIdentity, and that checks `isLocked`, actor, and status. This method would be similar to `isVisibilityAllowed`. Then if the constant is added as suggested above, that constant could also be private and just called with `self::`. This would allow combining logic in both places and ensuring check is consistent and easier to maintain in the future.

includes/Services/WikiRequestViewer.php
```diff
@@ -600,6 +613,22 @@ protected function submitForm(
 			return;
 		}
 
+		if ( isset( $formData['submit-abandon'] ) ) {
+			$abandonableStatuses = [ 'inreview', 'moredetails' ];
+			if ( $user->getActorId() !== $this->wikiRequestManager->getRequester()->getActorId() ||
+				!in_array( $this->wikiRequestManager->getStatus(), $abandonableStatuses, true )
+			) {
+				$out->addHTML( Html::errorBox( $this->context->msg( 'createwiki-error-cannot-abandon' )->escaped() ) );
```
Universal-Omega: Don't really need this message. Just the return is enough. In reality this should actually never be possible to hit this path.

Universal-Omega (changes requested):
A couple suggestions. On another note, I think we should also require a comment from a user abandoning the request that goes to the comments section, 1) so they don't accidentally abandon it and 2) so we know why they abandon it and that it was actually intentional.

We may want to add reopen conditions for this as well, and add extra checks in approved *and maybe* declined and other statuses so that a WR doesnt accidentally approve an abandoned request as well.

includes/Services/WikiRequestManager.php
```diff
@@ -407,6 +410,18 @@ public function getAllowedVisibilities( UserIdentity $user ): array {
 		return $allowedVisibilities;
 	}
 
+	public function canAbandonRequest( UserIdentity $user ): bool {
```
The-Voidwalker:
```suggestion
	public function canAbandonRequest( User $user ): bool {
```

`UserIdentity` does not include a definition for `getActorId`. If we need to use `UserIdentity` over `User`, then we should get the actor ID using `ActorNormalization`. e.g.:

```php
MediaWikiServices::getInstance()->getActorNormalization()->findActorId( $user, $dbw );
```

The-Voidwalker (changes requested):

Reception123:
The final patch has been successfully tested on Beta. The comments/suggestions have all been addressed and while it would've been ideal to also resolve the codecov CI test related issues, PRs have been approved in the past without them (https://github.com/miraheze/CreateWiki/pull/808#issuecomment-3970175148, https://github.com/miraheze/CreateWiki/pull/806#issuecomment-3916754363, https://github.com/miraheze/CreateWiki/pull/778#issuecomment-3314170009) and this PR is needed in order to better manage the wiki request queue and no longer require WRs to manually go through pending requests and decline them for no response (as the main script PR depends on this one)

PR #824, opened 2026-05-22
allow wiki requestors to abandon their request (T14115) - ci redo

Reception123:
Due to CI failing, I am redoing this PR.
Original PR is https://github.com/miraheze/CreateWiki/pull/818 and full credit goes to Void and UniversalOmega for their reviews/suggestions there.


Reception123:
looks like that won't work

PR #826, opened 2026-06-22
Add expiry option for inactive exempt wikis

Reception123:
In order to complete the transition away from using a manual Dormancy Policy exemption table, it's necessary to also include the option to set expiry as quite a few exemptions are given temporarily only. 

Note: This patch is partially written by LLMs with adjustments and has already been tested on beta.

https://github.com/miraheze/ManageWiki/pull/791 should also be merged once this is.


The-Voidwalker:
How are expired exemptions removed? I imagine we'd want that automatically handled by a script. Possibly at the start of manageInactiveWikis? May also be worth notifying wiki bureaucrats that the exemption has elapsed.

Reception123:
> How are expired exemptions removed? I imagine we'd want that automatically handled by a script. Possibly at the start of manageInactiveWikis? May also be worth notifying wiki bureaucrats that the exemption has elapsed.

Yeah, I was thinking a script would probably make sense for that.

Universal-Omega (approved):
As long as the script as discussed above is done also, this LGTM.

PR #831, opened 2026-07-19
popup after abandon; clarify that request is back in queue after NMD

Reception123:
Per T15678 and discussion in #tech on Discord

Note: This patch is partially written by LLMs with adjustments and has already been tested on beta.


includes/Services/WikiRequestViewer.php
```diff
@@ -604,6 +623,18 @@ protected function submitForm(
 			if ( $canEditReopen ) {
 				$this->wikiRequestManager->setStatus( 'inreview' );
 				$this->wikiRequestManager->log( $user, 'requestreopen' );
+
+				if ( $wasAwaitingMoreDetails ) {
+					$this->wikiRequestManager->addComment(
+						comment: $this->context->msg( 'requestwiki-moredetails-reopened' )
+							->inContentLanguage()->escaped(),
+						user: User::newSystemUser( 'CreateWiki Extension' ),
```
originalauthority: can use the same user instance as above?

Reception123: thanks, changed!

originalauthority:
LGTM!

PR #834, opened 2026-07-27
script to remove wikis whose exemptions have expired

Reception123:
For https://github.com/miraheze/CreateWiki/pull/826 per Void and CosmicAlpha. Not yet tested.


The-Voidwalker:
Should we also clear the expiry fields when removing the exemption? Not much of a reason to keep them, but also no particular need to remove them as far as I can see.

Reception123:
Thanks for the very quick review! And yeah, fair enough

Reception123:
@The-Voidwalker Would this be okay to merge now after the main PR is merged and CI is re-ran?

The-Voidwalker (approved):
Not that worried about code coverage here, looks good to me. Would prefer a dry run before a run with write enabled. Other than that though, I guess just remember to add a scheduled systemd job in puppet.

The-Voidwalker:
Oh, I guess it should have notifications similar to #842. Do wonder what that one is for though if this exists.

Universal-Omega (approved):

PR #842, opened 2026-08-08
handle removing inactive exempt expiry and send email notifications

Reception123:


Universal-Omega (changes requested):
When I mentioned a job, I didn't actually mean a job that does all wikis at once, but that does each wiki lazily, so it would only trigger on its own wiki, when it needs to, that way we don't need yet another periodic job, that runs continuously, nor do we need to query every single wiki at once, it would only do what it actually needs to do. It was however just a suggestion, a script may still have been more straightforward then this. But also, this job doesn't trigger anywhere at all, if what you want is to do like this job currently does and do all at once, a script may indeed be the better option though.

Reception123:
Per above, let's just go back to the script idea then.

PR #846, opened 2026-08-12
Allow disabling main page creation with a hook

lihaohong6:
Currently only supports disabling main page creation.

Can be modified to support other features (e.g. executing extra SQL statements).


lihaohong6:
Superseded by other workarounds.

PR #852, opened 2026-08-31
T15116: add a new wiki loadout selection feature

lihaohong6:


maintenance/ImportLoadoutXmlDump.php
```diff
@@ -0,0 +1,93 @@
...
+		$xmlPath = $this->getOption( 'xml' );
+
+		if ( !file_exists( $xmlPath ) || !is_readable( $xmlPath ) ) {
+			$this->fatalError( "XML dump file $xmlPath not found or not readable." );
+		}
+
+		$dbw = $this->getPrimaryDB();
+
+		try {
+			$result = Shell::makeScriptCommand(
```
Universal-Omega: I think we can use createChild here? But overall would it make more sense to use WikiImporterFactory here or not? maybe not unless this were made an actual job instead of a script where it is easier to call child scripts in scripts, but it also may be more useful to use WikiImporterFactory, not positive though.

Also benefit of WikiImporterFactory is we define a user so rather than using the default Maintenance user, we can use CreateWikiExtension. Also we should probably use username prefix, whether script is used or the service, for us that would be to the loadoutwiki for attribution, otherwise we need an explicit license on loadoutwiki not requiring attribution like that. For others that could be something else, so it should be a new configuration option.

maintenance/ImportLoadoutXmlDump.php
```diff
@@ -0,0 +1,93 @@
...
+		parent::__construct();
+
+		$this->addDescription( 'Imports a CreateWiki loadout XML dump into this wiki.' );
+		$this->addOption( 'xml', 'Path to the XML dump to import.', true, true );
+
+		$this->requireExtension( 'CreateWiki' );
+	}
+
+	public function execute(): void {
+		$logger = $this->getServiceContainer()->get( 'CreateWikiLogger' );
```
Universal-Omega: Please do it like:
https://github.com/miraheze/CreateWiki/blob/433a4650a4ff39759509600d725b4fcef5d36c64/maintenance/DeleteWiki.php#L23-L26.

phan-var can be dropped below this if done like that also.

maintenance/ImportLoadoutXmlDump.php
```diff
@@ -0,0 +1,93 @@
...
+
+			$this->logger->info( 'Loadout import finished importing the XML dump.' );
+
+			$siteStatsInit = new SiteStatsInit();
+			$siteStatsInit->refresh();
+
+			SiteStatsUpdate::cacheUpdate( $dbw );
+			$this->logger->info( 'Loadout import finished updateSiteStats.' );
+
+			$maintenance = new FakeMaintenance;
```
Universal-Omega: We don't need FakeMaintenance here, as this itself is maint, we can use $this->createChild.

maintenance/ImportLoadoutXmlDump.php
```diff
@@ -0,0 +1,93 @@
...
+				'walltime' => 0,
+			] )->execute();
+
+			if ( $result->getExitCode() !== 0 ) {
+				$this->fatalError( 'Loadout import failed to import the dump file: ' . $result->getStderr() );
+			}
+
+			$this->logger->info( 'Loadout import finished importing the XML dump.' );
+
+			$siteStatsInit = new SiteStatsInit();
```
Universal-Omega: Probably a left over from ImportDump, for this one didn't this is a script itself, we can probably use createChild( InitSiteStats::class ) with options for this and below SiteStatsUpdate stuff. If we do that, we can also remove $dbw above as well.

includes/Loadout/LoadoutFormBuilder.php
```diff
@@ -0,0 +1,77 @@
...
+class LoadoutFormBuilder {
+
+	public function __construct(
+		private readonly LoadoutManager $loadoutManager,
+		private readonly MessageLocalizer $messageLocalizer,
+	) {
+	}
+
+	public function isEnabled(): bool {
+		return $this->loadoutManager->isEnabled();
```
Universal-Omega: This seems kinda redundant to me? I think we can just call this directly rather than another method for it here?

includes/Loadout/LoadoutManager.php
```diff
@@ -0,0 +1,187 @@
...
+				'XML dump file {path} not found or not readable for wiki {dbname}',
+				[
+					'dbname' => $dbname,
+					'path' => $xmlPath,
+				]
+			);
+			return;
+		}
+
+		// The import needs to be done in a separate process because the
```
Universal-Omega, line 133: Sounds like a recipe for a race condition?

maintenance/ImportLoadoutXmlDump.php
```diff
@@ -0,0 +1,93 @@
...
+			$siteStatsInit->refresh();
+
+			SiteStatsUpdate::cacheUpdate( $dbw );
+			$this->logger->info( 'Loadout import finished updateSiteStats.' );
+
+			$maintenance = new FakeMaintenance;
+			$rebuildText = $maintenance->createChild( RebuildTextIndex::class );
```
Universal-Omega: Should check for MainConfigNames::DisableSearchUpdate

maintenance/ImportLoadoutXmlDump.php
```diff
@@ -0,0 +1,93 @@
...
+	public function execute(): void {
+		$logger = $this->getServiceContainer()->get( 'CreateWikiLogger' );
+		'@phan-var LoggerInterface $logger';
+		$this->logger = $logger;
+		$this->logger->info( 'Loadout import started.' );
+
+		$xmlPath = $this->getOption( 'xml' );
+
+		if ( !file_exists( $xmlPath ) || !is_readable( $xmlPath ) ) {
+			$this->fatalError( "XML dump file $xmlPath not found or not readable." );
```
Universal-Omega: 1. Not positive if fatalError is best here or not, probably fine though
2. Above it should also log to CreateWikiLogger though otherwise that isn't really logged other than to stdout

maintenance/ImportLoadoutXmlDump.php
```diff
@@ -0,0 +1,93 @@
...
+		$this->addOption( 'xml', 'Path to the XML dump to import.', true, true );
+
+		$this->requireExtension( 'CreateWiki' );
+	}
+
+	public function execute(): void {
+		$logger = $this->getServiceContainer()->get( 'CreateWikiLogger' );
+		'@phan-var LoggerInterface $logger';
+		$this->logger = $logger;
+		$this->logger->info( 'Loadout import started.' );
```
Universal-Omega: For this and other logs, maybe add what wiki?

Universal-Omega (changes requested):
An initial review but could still be a little more review later.

includes/Loadout/LoadoutManager.php
lihaohong6: This is the same pattern as `PopulateMainPage` (https://github.com/miraheze/CreateWiki/blob/433a4650a4ff39759509600d725b4fcef5d36c64/includes/Services/WikiManagerFactory.php#L277) and I think as long as it runs in the deferred update the list of wikis should be up-to-date.

With that said maybe the occasional missing main page is due to this race condition.

Universal-Omega: Yeah I didn't mean it was different just that it would introduce a new race condition also.

Universal-Omega: I actually think this may be the opposite, run this in the same thread, after first regenerating database lists, and it may resolve that race condition.

maintenance/ImportLoadoutXmlDump.php
lihaohong6: An old version uses WikiImporterFactory but the import was super slow for some reason (https://github.com/miraheze/CreateWikiLoadout/blob/6f41cfb5819a32f321cfa57b6faaa71a8155eed2/maintenance/ImportLoadoutXmlDump.php#L59). Core's importDump turned out to be a lot faster. As for the user, since we control the XML dump we can just make the user whichever one we'd like? I'm not sure how the local user creation process works though.

As for attribution, I think we don't need the full history since we can just make the edit summary something like `Imported from loadout wiki. See [loadout wiki link] for attribution`.

includes/Loadout/LoadoutFormBuilder.php
```diff
@@ -0,0 +1,73 @@
...
+			'type' => 'select',
+			'label-message' => 'createwiki-label-loadout',
+			'help-raw' => $this->buildHelpText( $options ),
+			'options' => $options,
+			'validation-callback' => [ $this, 'validateLoadout' ],
+			'default' => '',
+		];
+	}
+
+	public function validateLoadout( ?string $loadout ): bool|Message {
```
Universal-Omega:
```suggestion
	public function validateLoadout( ?string $loadout ): Message|true {
```

includes/Loadout/LoadoutFormBuilder.php
```diff
@@ -0,0 +1,73 @@
...
+		return $options;
+	}
+
+	/**
+	 * @param array<string,string> $options
+	 */
+	private function buildHelpText( array $options ): string {
+		$items = '';
+		foreach ( $options as $label => $loadout ) {
+			$messageKey = 'createwiki-help-loadout-' . ( $loadout === '' ? 'none' : $loadout );
```
Universal-Omega:
```suggestion
			$messageKey = 'createwiki-help-loadout-' . ( $loadout ?: 'none' );
```

includes/Loadout/LoadoutFormBuilder.php
```diff
@@ -0,0 +1,73 @@
...
+	) {
+	}
+
+	public function getFormDescriptor(): array {
+		$options = $this->buildOptions();
+		return [
+			'type' => 'select',
+			'label-message' => 'createwiki-label-loadout',
+			'help-raw' => $this->buildHelpText( $options ),
+			'options' => $options,
```
Universal-Omega: We can use `options-messages` instead here.

includes/Loadout/LoadoutManager.php
lihaohong6: The database list is already regenerated before the script run. Before `onCreateWikiAfterCreationWithExtraData` there's a `resetDatabaseLists` call which should regenerate the database list. At this point the current process still has stale DB data but any child process that's spawned will read the updated DB list. It might be possible to nuke the cache on the current process but I don't think it'll be easy since everything is already initialized with stale data.

includes/Loadout/LoadoutFormBuilder.php
```diff
@@ -0,0 +1,73 @@
...
+		private readonly MessageLocalizer $messageLocalizer,
+	) {
+	}
+
+	public function getFormDescriptor(): array {
+		$options = $this->buildOptions();
+		return [
+			'type' => 'select',
+			'label-message' => 'createwiki-label-loadout',
+			'help-raw' => $this->buildHelpText( $options ),
```
Universal-Omega: I think we may be able to use `help-messages` here which creates messages from keys and does ->parse, we could then either do the dom building via the message itself or CSS by adding a `csshelpclass` here maybe? I think that may significantly help us clean up some of the raw element creation here also. Not 100% positive here though, but if we can that would much better I think.

includes/Loadout/LoadoutFormBuilder.php
```diff
@@ -0,0 +1,73 @@
...
+
+	public function __construct(
+		private readonly LoadoutManager $loadoutManager,
+		private readonly MessageLocalizer $messageLocalizer,
+	) {
+	}
+
+	public function getFormDescriptor(): array {
+		$options = $this->buildOptions();
+		return [
```
Universal-Omega, line 18: Needs `'cssclass' => 'ext-createwiki-infuse',`

includes/Loadout/LoadoutManager.php
```diff
@@ -0,0 +1,187 @@
...
+					'dbname' => $dbname,
+					'error' => $result->getStderr(),
+				]
+			);
+		}
+	}
+
+	private function getModuleFactory( string $dbname ): ?ModuleFactory {
+		if ( $this->moduleFactory === null ) {
+			$this->logger->error(
```
Universal-Omega: This log is honestly not all that helpful and if removed, since moduleFactory itself is now just null when it doesn't exist, this entire method could be removed and just checked the property directly to simplify it.

Universal-Omega (changes requested):
A little more review. Overall looks pretty good, just some things. A lot of it kinda minor.

extension.json
```diff
@@ -379,6 +400,14 @@
 			"description": "Array. Possible options for marking a wiki as exempt from inactivity.",
 			"value": []
 		},
+		"CreateWikiLoadoutEnabled": {
```
Universal-Omega: Hmm.. maybe we could just remove this config, and use CreateWikiLoadoutConfigs as the source of truth for enabled also. IE empty is disabled?

includes/Loadout/LoadoutFormBuilder.php
lihaohong6: I changed things around. The downside is that we no longer get all the `<li>` entries inside a single `<ul>` since each message is parsed separately.

includes/Loadout/LoadoutFormBuilder.php
```diff
@@ -38,36 +36,34 @@ public function validateLoadout( ?string $loadout ): Message|true {
...
-				Html::element( 'strong', [], $label ),
-				$this->messageLocalizer->msg( 'colon-separator' )->escaped(),
-				$this->messageLocalizer->msg( $messageKey )->parse(),
-			] ) );
+	private function buildHelpMessages(): array {
+		$messages = [ 'createwiki-help-loadout' ];
+		foreach ( [ 'none', ...$this->loadoutManager->getLoadoutNames() ] as $loadout ) {
+			$messages[] = [
+				'createwiki-help-loadout-item',
+				$this->messageLocalizer->msg( "createwiki-label-loadout-$loadout" ),
```
Universal-Omega: This can be simplified more *I think*:
1. Just do  `$messages[] = [ 'createwiki-help-loadout-item', $loadout ]`
2. Update the message to do `* '''{{int:createwiki-label-loadout-$1}}''': {{int:createwiki-help-loadout-$1}}`

I am not positive, so it needs testing, also not positive if it has negative performance parsing there instead.

includes/Loadout/LoadoutFormBuilder.php
```diff
@@ -0,0 +1,69 @@
...
+			'cssclass' => 'ext-createwiki-infuse',
+			'label-message' => 'createwiki-label-loadout',
+			'help-messages' => $this->buildHelpMessages(),
+			'options-messages' => $this->buildOptionsMessages(),
+			'validation-callback' => [ $this, 'validateLoadout' ],
+			'default' => '',
+		];
+	}
+
+	public function validateLoadout( ?string $loadout ): Message|true {
```
Universal-Omega, line 29: This can probably move to CreateWikiValidator, alongside isEnabled -> isLoadoutEnabled or something, and getLoadoutNames then inject CreateWikiValidator. Only do this if doing the comment I mentioned below also, because my main goal is to reduce what we need to pass, not expand it, if this is done:
1. No more need to inject MessageLocalizer, or LoadoutManager here (just CreateWikiValidator)
2. Can remove getLoadoutNames and isEnabled from LoadoutManager, no replacement needed, they aren't used there. Can also avoid the need for ServiceOptions there at all if we other helpers like getLoadoutConfig( $loadoutName ), loadoutExists( $loadoutName )... might be overall cleaner but would then replace ServiceOptions with another injection. Might be worth it to reduce some duplication and testability though.
3. Remove LoadoutManager injection from SpecialCreateWiki entirely. It already has CreateWikiValidator. Could also remove the import if FIELD_NAME is moved to this class (which kinda makes more sense to me actually)
4. The same can be done for the Loadout HookHandler if you remove onCreateWikiAfterCreationWithExtraData call, and apply inline, which should work, and just use validator in the handle. This has another side effect though of a new injection here for validator.

Universal-Omega: Feel free to completely ignore this. A lot of it is complicated, I don't think we should yet but for a future cleanup later which is why I still posted this.

includes/Loadout/LoadoutManager.php
```diff
@@ -0,0 +1,171 @@
...
+	public function getLoadoutNames(): array {
+		$loadouts = $this->options->get( ConfigNames::LoadoutConfigs );
+		'@phan-var array<string, array> $loadouts';
+		return array_keys( $loadouts );
+	}
+
+	/**
+	 * Whether the request's loadout imports an XML dump, which provides a custom main page.
+	 */
+	public function willImportXml( array $extra ): bool {
```
Universal-Omega, line 49: I am little worried about this. I don't believe we should have to keep the XML on the disk permanently, because we use swift for file storage. My thought is we do like we do for the persistent model:

See:
https://github.com/miraheze/CreateWiki/blob/433a4650a4ff39759509600d725b4fcef5d36c64/includes/Jobs/RequestWikiAIJob.php#L42-L51
https://github.com/miraheze/CreateWiki/blob/433a4650a4ff39759509600d725b4fcef5d36c64/maintenance/CreatePersistentModel.php#L94-L99

And also this from MirahezeMagic, no longer exists because we no longer use it but the idea remains and it did work:
https://github.com/miraheze/MirahezeMagic/blob/0c2235fdde9a958ab0219f0beeee5581781861cd/includes/MirahezeMagicHooks.php#L256-L278

We should also probably add a specific script to generate the XML dump specifically from the LoadConfigs option, where it gets what wiki to generate there, with options to provide users, attribution, etc.. as needed to create the XML we need.

Wondering your thoughts though?

Universal-Omega (commented):
Just some comments, not really requesting changes yet for these, though the one about the message is probably the one one immediately actionable if it works properly.

includes/Loadout/LoadoutFormBuilder.php
lihaohong6: Changed. The time spent parsing this is probably negligible since it's only a few messages anyway.

includes/Loadout/LoadoutManager.php
lihaohong6: I think the newest commit addresses this in a way that doesn't introduce any configuration or hook changes. The importer now supports `mwstore` files which can point to a Swift container path, and there is another maintenance script `StoreLoadoutXmlDump` that saves a local XML dump to Swift. The maintenance script does the minimal thing so that we can start with manually generating the XML dump and saving it to Swift. Once the export format is determined we can start automating XML dump creation.

